PR #6, opened 2021-02-19
chore: use semver to determine API compatibility

### LINKED ISSUE #5: Use semver to determine API compatibility
The API compatibility version should be replaced with a true semver check. This will be a port of https://github.com/open-telemetry/opentelemetry-js/pull/1772

dyladan:
Fixes #5


dyladan:
@MSNev the biggest changes here are related to the diag logger because it behaves differently than the other APIs so I would appreciated a peek

src/api/diag.ts
```diff
@@ -22,63 +22,38 @@ import {
...
-    let _logLevel: DiagLogLevel = DiagLogLevel.INFO;
-    let _filteredLogger: DiagLogger | null;
-    let _logger: DiagLogger = createNoopDiagLogger();
+    const _noopLogger = createNoopDiagLogger();
 
     function _logProxy(funcName: keyof DiagLogger): DiagLogFunction {
       return function () {
         const orgArguments = arguments as unknown;
-        const theLogger = _filteredLogger || _logger;
+        const theLogger = getGlobal('filteredDiagLogger') || _noopLogger;
```
MSNev, line 56: Why register the filteredDiagLogger and diagLogger as globals, they are only used by the diagAPI and previously included as embedded local vars for the instance.

I don't see why we would need to store the globals diagLogger, filteredDiagLogger and diagApi?

So I would suggest either keeping the diagLogger and filteredDiagLogger as locals or 
Don't register the DiagAPI as a global as this change basically makes the need to have it as a global to disappear

MSNev: Maybe another way to say this would be

diagLogger === DiagApi.diagLogger
filterDiagLogger ===  DiagApi.filteredDiagLogger === wrapped(DiagApi.diagLogger) === wrapped(diagLogger)

src/api/diag.ts
```diff
@@ -89,34 +64,44 @@ export class DiagAPI implements DiagLogger {
       };
     }
 
+    function getLevel(): DiagLogLevel {
+      return getGlobal('diagLogLevel') ?? DiagLogLevel.INFO;
```
MSNev, line 68: Likewise for this, why does it need to be global as well as the diagApi?

src/api/diag.ts
```diff
@@ -22,63 +22,38 @@ import {
...
 import { DiagLogLevel, createLogLevelDiagLogger } from '../diag/logLevel';
 import {
-  API_BACKWARDS_COMPATIBILITY_VERSION,
-  GLOBAL_DIAG_LOGGER_API_KEY,
-  makeGetter,
-  _global,
-} from './global-utils';
-
-/** Internal simple Noop Diag API that returns a noop logger and does not allow any changes */
-function noopDiagApi(): DiagAPI {
```
MSNev, line 32: This was purely for duplicating the previous globals, so when it didn't match it returned a noop version of the requested global -- definitely not sad to remove this as didn't like this pattern.

src/api/diag.ts
dyladan: The logging needs to be global because if the user sets the log level of the API they installed for the application it needs to be applied to other API modules installed in the deeply nested node_modules structure.

MSNev: Which is the same issue and why the DiagAPI is global -- so if they always get back the same DiagAPI object with these embedded it achieves the same goal.

Another issue would be that *if* the implementation of filtered / diag changes in some incompatible way, e.g. new code wants new functions but old version registered then this would cause a problem for the new user. This is also avoided by keeping the logger instances within the global DiagApi (mostly) -- there would still be the issue that the global DiagApi would also have the same issue.

dyladan: >There would still be the issue that the global DiagApi would also have the same issue.

Exactly. We need to have some level of guaranteed compatibility if we want to share anything. We _could_ just make the level global and make everything else module-global but that would require another rearchitecting of the diag logger in general to allow the log level to be changed on the fly

MSNev: > some level of guaranteed compatibility

Agree, I was trying to provide a simple hypothetical. As if we only have the diag Api global, which internally uses loggers then we only have 1 place to worry about compatibility issues (again mostly, based on the assumption that DiagLoggers should only be injected into the diagAPI)

> We could just make the level global

With the level, embedded in the global DiagAPI, it effectively would be...

And in addition to that, we already have the global environment (in the SDK) which uses (or should use) the runtime environment to set this value during initialization, so under normal conditions it will always be set to the same value. And as this is for Diagnostic tracing only there is now real need (exception below) to have different levels.

> and make everything else module-global

I don't think we need to go that far, the namespace idea from Bart would work it those cases. And we could build / extend the namespace to allow setting it own local log level, which is part of the reason for exposing the getLogger() on the api to allow it to bypass the global log level if needed.

I went through the same thoughts that I think you are having now when I decided to only make the DiagAPI global and embed the logger and level into the API only. 

Specifically, thinking around the use case when using self contained modules that don't use globals as each module effectively gets own self contained local instances (in their own module closures), even when running in the same run time environment. These cases are handled by having the global root objects, and by having everything in the single object is saves the need to populate and lookup the other "effectively" internal properties. 

There is however, a bigger (spec) issue with having the global in general. But that a discussion for another day and is generally not an issue for server side (nodejs) usages.

MSNev: I will add a little bit to this based on additional thoughts after posting above

> and make everything else module-global

I agree that we did lose (remove) this capability by removing the xxxxFromConfig() helper and the ability to pass / set the logger instance via config per component. And if we want it back we would either need to build / enhance using the namespace approach or allow it via config again (I don't want the OptionalDiagLogger type for passing as arguments back as that was purely for compatibility and passing a "context" is better than having lots of arguments.)

dyladan: All three (logger, logLevel, and filteredLogger) need to be global in order to allow changing log level and logging destination.

If level is changed, a new filtered logger needs be be created from the old logger and new level.

If logger is changed, a new filtered logger needs to be created from the new logger and old level.

Filtered logger needs to be global so that it can be used in many places.

I think you misunderstood what I meant by module global. If an application depends on the API and also depends on some dependency that also depends on the API, if those API versions don't exactly match there will be a duplicated installation of the API in the node modules folder

```
node_modules
	@opentelemetry/api

	some-dependency
		@opentelemetry-api
```

Anything that is in the singleton of one install will not be available by the singleton of the other install without using the `global` object to communicate (global in node, window in browser).

dyladan: We can get around having 3 globals by only making the filtered logger global and having the `set` method set both the logger and the level. This simplifies the code but means you can't change log level on the fly. The way you've created the logger already makes on-the-fly log level changing somewhat difficult so maybe that's not such a problem.

MSNev: > I think you misunderstood what I meant by module global

No, I completely understand and this is exactly the same issue for trace, context and propagation. Which is also why I copied the same pattern as we only have 1 pattern to "fix" (i.e. the returning of Noop instances if the version of the API doesn't match)

> We can get around having 3 globals by only making the filtered logger global

I still don't think so, because the filtered logger delegates to the underlying logger.

Notwithstanding the component config logger that got removed, I *think* your trying to work around another underlying issue, that's not directly related to the DiagAPI, where the way ```api.diag``` is used is highlighting the situation and I think a better solution (rather than this "work around") would be option 3 below

So I still think having the DiagAPI as the global is the correct option. With the possible changes required would be either :-
- keep existing, Api Version mismatch == Noop for 2nd caller to "register" the global (not ideal -- would require additional documentation to call this out)
- Change so each version gets it's own DiagAPI instance and doesn't share the logger -- really same situation as returning a Noop, but if we have the SDK initialize the properties then all "version" would get their own logger of the same type and level. This "works" for the simple loggers (Console, Noop) we have today, but could complicate issues for File based etc.
- Change the "global" fetching to use a component specific (sub-version) (i.e. "DiagAPI version") rather than the global API version to return either of the above.
  - So the issue of Api v1.0 and v1.1 doesn't trigger the issue
  - And it would only occur if we introduce a breaking DiagAPI version change
  - Or we have the ability to say that the Noop fallback would only be triggered for some components on major upgrade (v1 -> v2)
- Or a complete alternative -- extend the withNameSpace() idea to handle level and namespace. 
  - For this it would be the underlying ```DiagLogger``` instance that would be global and change so it doesn't fallback to Noop instance based on the version above (option 3)
  - Each "version" of the API would create a new ```api.diag``` instance and logging level (Using existing level logic)
  - Assumes that setting the logger on one instance changes the logger for all (but that assumption is also coming from your change to make everything global) -- but not the level.
  - So this could be a new function on DiagAPI something like ```with(ns: string, level?: DiagLogLevel): DiagLogger```
    - ```api.diag.with("myNamespace")``` or ```api.diag.with("myNameSpace", DiagLogLevel.DEBUG)```

> This simplifies the code but means you can't change log level on the fly.

As a diagnostic logging system, I don't think we really need to support this -- do we? I can think of some debugging cases where it would be "nice" to change it on the fly, but if we have a single global (version shared -- option 3) instance then this would be supported and if we keep the existing Noop we just say it's not supported.

dyladan: >As a diagnostic logging system, I don't think we really need to support this -- do we? I can think of some debugging cases where it would be "nice" to change it on the fly, but if we have a single global (version shared -- option 3) instance then this would be supported and if we keep the existing Noop we just say it's not supported.

I tend to agree here. I think the diagnostic logging is likely to be left turned off in almost all production cases and this would be something local developers would temporarily enable (or in dev/staging) and a process restart isn't a huge burden for that (or `api.diag.disable(); api.diag.setLogger()`).

>Change the "global" fetching to use a component specific (sub-version) (i.e. "DiagAPI version") rather than the global API version to return either of the above.

We considered this, but signals already interact with each other and allowing all the mixing and matching is a potential point for trouble. The only place where APIs should be registered is by the end user anyways, so it should be the case that only a single API instance ever registers globals. I would almost say we should do the opposite and make the entire API a single global instance instead of managing multiple globals. The only reason I didn't do that is because this simplified the implementation.

dyladan:
@MSNev While trying to get together an implementation that would not allow changing log levels in the global logger, I came across the test titled `'diag setLogLevel is ignored when using a specific logger'`. It seems to me if the global log level is `INFO` and i create a filtered logger with the `DEBUG` level. I should be able to see `debug` level logs when I call it. But this test appears to assert the opposite? What is the point of creating a separate logger if not to get more specific logging from one component than what the global logger is set to?

MSNev:
> @MSNev Nev Wylie FTE While trying to get together an implementation that would not allow changing log levels in the global logger, I came across the test titled `'diag setLogLevel is ignored when using a specific logger'`. It seems to me if the global log level is `INFO` and i create a filtered logger with the `DEBUG` level. I should be able to see `debug` level logs when I call it. But this test appears to assert the opposite? What is the point of creating a separate logger if not to get more specific logging from one component than what the global logger is set to?
> 
> edit: I resolved it by doing this. Can you please take a look? [dyladan#1](https://github.com/dyladan/opentelemetry-js-api/pull/1)

The test is asserting that if you just create a new level of "X" but the user / application set the level higher globally AND you have created a filtered logger using the default (without specifying a logger), it will default to delegating the logging api.diag, which will then obey the global logging level request. And if you want to override the global level then you need to specify the specific logger, so you need to get the underlying (unfiltered logger) via ```api.diag.getLogger()```

This is documented explicitly in the TS doc of the createLogLevelDIagLogger()
https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-api/src/diag/logLevel.ts#L67-L79

dyladan:
> > @MSNev Nev Wylie FTE While trying to get together an implementation that would not allow changing log levels in the global logger, I came across the test titled `'diag setLogLevel is ignored when using a specific logger'`. It seems to me if the global log level is `INFO` and i create a filtered logger with the `DEBUG` level. I should be able to see `debug` level logs when I call it. But this test appears to assert the opposite? What is the point of creating a separate logger if not to get more specific logging from one component than what the global logger is set to?
> > edit: I resolved it by doing this. Can you please take a look? [dyladan#1](https://github.com/dyladan/opentelemetry-js-api/pull/1)
> 
> The test is asserting that if you just create a new level of "X" but the user / application set the level higher globally AND you have created a filtered logger using the default (without specifying a logger), it will default to delegating the logging api.diag, which will then obey the global logging level request. And if you want to override the global level then you need to specify the specific logger, so you need to get the underlying (unfiltered logger) via `api.diag.getLogger()`
> 
> This is documented explicitly in the TS doc of the createLogLevelDIagLogger()
> [https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-api/src/diag/logLevel.ts#L67-L79](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-api/src/diag/logLevel.ts?rgh-link-date=2021-02-23T19%3A10%3A26Z#L67-L79)

That's quite confusing to me. I would expect that if I called `createLogLevelDiagLogger` with a level the returned logger would respect that level. With no logger passed, I would expect it to use the same destination as the global, but not necessarily respect the global level.

MSNev:
> That's quite confusing to me. I would expect that if I called `createLogLevelDiagLogger` with a level the returned logger would respect that level. With no logger passed, I would expect it to use the same destination as the global, but not necessarily respect the global level.

Yes, I agree. It came about from a comment from Bart where *if* the user has set the global logging level to NONE then he did not expect that we (by default) would override their level request. But we still had code (now removed as part of the config) that wanted to override and set directly.

dyladan:
@obecny do you mind if we reverse this behavior?

obecny:
> > That's quite confusing to me. I would expect that if I called `createLogLevelDiagLogger` with a level the returned logger would respect that level. With no logger passed, I would expect it to use the same destination as the global, but not necessarily respect the global level.
> 
> Yes, I agree. It came about from a comment from Bart where _if_ the user has set the global logging level to NONE then he did not expect that we (by default) would override their level request. But we still had code (now removed as part of the config) that wanted to override and set directly.

which comment ?

dyladan:
> which comment ?

I said:

>I would expect that if I called createLogLevelDiagLogger with a level the returned logger would respect that level. With no logger passed, I would expect it to use the same destination as the global, but not necessarily respect the global level.

Nev said:

>Yes, I agree. It came about from a comment from Bart where if the user has set the global logging level to NONE then he did not expect that we (by default) would override their level request. But we still had code (now removed as part of the config) that wanted to override and set directly.

Would you be ok if that was the behavior?

obecny:
> > which comment ?
> 
> I said:
> 
> > I would expect that if I called createLogLevelDiagLogger with a level the returned logger would respect that level. With no logger passed, I would expect it to use the same destination as the global, but not necessarily respect the global level.
> 
> Nev said:
> 
> > Yes, I agree. It came about from a comment from Bart where if the user has set the global logging level to NONE then he did not expect that we (by default) would override their level request. But we still had code (now removed as part of the config) that wanted to override and set directly.
> 
> Would you be ok if that was the behavior?

@dyladan I would expect the behaviour you exactly described, if you pass some params to a function it should respect this params, otherwise it doesn't make sense to insert any params......, not sure why it would be different. But then @MSNev said that it is working differently because of my comment, so @MSNev which comment I did that it is working differently ???

MSNev:
I found the original comment and sorry @obecny I miss-attributed (and slightly miss-remembered) it to you when it was actually @vmarchaud on this comment. 

The comment was around having the "forcedInfo" and specifically this bit

> i don't expect having log coming where i explicitly dont want them

https://github.com/open-telemetry/opentelemetry-js/pull/1880#discussion_r573183063

dyladan:
@vmarchaud do you have an opinion on this?

vmarchaud:
> The comment was around having the "forcedInfo" and specifically this bit
>> i don't expect having log coming where i explicitly dont want them

I was refering to the fact that we couldn't (from my understand of the proposal) silence "forcedInfo".
If i get correctly the problem here, we are talking about the situation where i would have set the level to INFO on the global diag api but someone created another logger with a different level, in this case should we respect the global level or the level provided when constructing the logger ?
I believe the correct way (and whats possible with other APIs) is to use the level provided by the constructor, so i agree with @dyladan.

I understand why it oppose with my previous comment, but i think we should never construct a diag logger in any sdk components, only users should use this, so we can't really have a situation where the logger is using a level that wasn't choosen by the user

MSNev:
> but i think we should never construct a diag logger in any sdk components, only users should use this, so we can't really have a situation where the logger is using a level that wasn't choosen by the user

I TOTALLY agree with this, having the constructors / helpers is a "nice' to have for SDK / component developers during development ONLY.

src/api/diag.ts
```diff
@@ -134,17 +103,22 @@ export class DiagAPI implements DiagLogger {
...
   public info!: DiagLogFunction;
   public warn!: DiagLogFunction;
   public error!: DiagLogFunction;
+
+  /**
+   * Unregister the global logger and return to Noop
+   */
+  public disable!: () => void;
+
+  public getLoggingDestination!: () => DiagLogger;
```
MSNev, line 123: The name is a bit "wordy" (i.e. minification ;-) ) and it's not really the "destination".

Perhaps ```getRootLogger```?

dyladan: `childLogger`?

src/api/diag.ts
```diff
@@ -95,28 +70,22 @@ export class DiagAPI implements DiagLogger {
...
+    self.getLoggingDestination = () => {
+      return getGlobal('diag')?.getLoggingDestination() ?? _noopLogger;
+    };
 
-      return prevLogger;
+    self.setLogger = (
+      logger: DiagLogger = _noopLogger,
+      logLevel: DiagLogLevel = DiagLogLevel.INFO
+    ) => {
+      registerGlobal('diag', createLogLevelDiagLogger(logLevel, logger));
```
MSNev, line 84: I had the simple "check" here for logger !== self, just in case someone does something stupid as it will cause a stack overflow internally in our code if they do this ```api.diag.setLogger(api.diag, DiagLoggerLevel.DEBUG)```

Which is now the only way they can set the logging level programatically which affecting the logger. And yes I know they *should* use ```api.diag.setLogger(api.diag.getLoggingDestination(), DiagLoggerLevel.DEBUG))``` But this is also why it was a separate function (along with the previous no defunct usages)

src/api/diag.ts
MSNev: that works too

obecny: The micro optimisation which causes the descriptive name to be changed to something less meaningful is just wrong. This is the last thing the dev should be worried about. This is a job for obfuscators and minifications process, including the google closure with advanced option for the whole bundle where such name can be changed to `ab` for example. The name should be descriptive not minified. If this function returns something then starting with the prefix `get` is self explanatory already.

src/api/diag.ts
dyladan: 28b2f3e2d7fed89105f52b18abd00d0716558361

src/api/diag.ts
dyladan: I think `getChildLogger` is still descriptive and possibly more accurate than `getLoggingDestination`. The child may be a logging destination, or some other type of filter, or another logging framework like bunyan or winston.

.eslintrc.js
```diff
@@ -44,6 +44,7 @@ module.exports = {
             "files": ["test/**/*.ts"],
             "rules": {
                 "no-empty": "off",
+                "@typescript-eslint/ban-ts-comment": "off",
```
dyladan, line 47: We need ts comments for asserting a type check will fail. This allows it in tests only

src/api/diag.ts
obecny: I was mainly referring to `minification` argument as the reason for shortening the name and not really the name itself, which is often the hardest part in coding anyway ;)

dyladan: Are you ok with the name `getChildLogger`?

obecny: > Are you ok with the name `getChildLogger`?

of course

dyladan:
@MSNev I marked our previous conversation around globals as resolved because it was a long conversation largely about code that has been removed. If you still feel my use of globals is invalid please open a new comment thread. Otherwise I would appreciate another review.

src/api/diag.ts
MSNev: Ok, I'll try not to mention ```minification``` quite so much 😅. 

But as JS developers, especially for API's / SDK's for browser (client) based packages where the total package size and PLT is a concern. We need to be aware and not design something that would cause someone to choose not to use OT because the **commonly used API surface** being too extremely verbose when it doesn't need to be. i.e. We don't want to self obfuscate the name of the function, but if we can describe the operation / property more simply then we should. Because I total agree with your following comment 

> The micro optimisation which causes the descriptive name to be changed to something less meaningful is just wrong.


But,

> This is a job for obfuscators and minifications process

Not always, generally public names of classes / objects (especially for API's) cannot be minified to ensure cross communication. 

And anyone using an obfuscator for their entire codebase then yes it doesn't matter in any way.

src/api/diag.ts
```diff
@@ -95,28 +70,23 @@ export class DiagAPI implements DiagLogger {
     // DiagAPI specific functions
 
     self.getLogger = (): DiagLogger => {
-      // Return itself if no existing logger is defined (defaults effectively to a Noop)
-      return _logger;
+      return getGlobal('diag') || _noopLogger;
```
MSNev, line 73: Thought: Do we even need this anymore?

Can we just have this as the getChildLogger(), as there is no need to call getLogger() to bypass the filtering (as it won't) and the DiagAPI is a logger so you don't need to call ```api.diag.getLogger().debug(...)```

MSNev: Especially, as this getLogger() was previously just here to return the child logger.

src/diag/logger.ts
```diff
@@ -59,6 +59,10 @@ export interface DiagLogger {
   verbose: DiagLogFunction;
 }
 
+export interface FilteredDiagLogger extends DiagLogger {
```
MSNev, line 62: Question: Do we really need a new interface for this?

What about just adding this as an optional function, while it makes the usage a little more tricky we have to do this now *if* we don't already know (because of type enforcement) that this isn't a filtered logger. And the usages of this should be minimal anyway.

Would also suggest that if you want to keep it the name might be better as ```ParentDiagLogger``` with it's function of ```getChildLogger```

src/api/diag.ts
MSNev: And this would mean that the usage on L56 would have to change.

src/api/diag.ts
```diff
@@ -95,28 +70,23 @@ export class DiagAPI implements DiagLogger {
...
+      return getGlobal('diag')?.getLoggingDestination() ?? _noopLogger;
+    };
 
-      return prevLogger;
+    self.setLogger = (
+      logger: DiagLogger = _noopLogger,
+      logLevel: DiagLogLevel = DiagLogLevel.INFO
+    ) => {
+      logger = logger === self ? self.getLoggingDestination() : logger;
+      registerGlobal('diag', createLogLevelDiagLogger(logLevel, logger), true);
```
MSNev, line 82: Might be worth adding a comment that we are allowing the override of the default to both replacing any existing logger AND to avoid any possible endless loop (because of someone refactoring the code as the existing implementation doesn't have an issue) -- just so it make anyone touching the code in the future to think about that scenario before just blindly changing it.

src/api/diag.ts
dyladan: Do we need the diag API to understand the concept of filtering at all? We could just have the `DiagLogger` interface, and require the user to construct a valid logger for their use case like `diag.setLogger(logLevelFilteredLogger(level, console))`

test/diag/logLevel.test.ts
```diff
@@ -50,8 +50,7 @@ describe('LogLevelFilter DiagLogger', () => {
 
   beforeEach(() => {
     // Set no logger so that sinon doesn't complain about TypeError: Attempted to wrap xxxx which is already wrapped
-    diag.setLogger();
-    diag.setLogLevel(DiagLogLevel.INFO);
+    diag.disable();
```
MSNev, line 53: You may still need to reset the global log level as part of this as I found when running tests locally and in the CI environment that the order of execution can pollute other tests when they don't manager (or expect) the logging level to have changed.

But I think this is clearer than the side-effect version of setting the logger to undefined.

src/api/diag.ts
MSNev: The approach I was taking is that users and creators of loggers shouldn't need to worry about log level filtering as this should be controlled by the DiagAPI and any initialization configuration (as it *should* only be set once).

So I prefer what you have done in the setLogger ```registerGlobal('diag', createLogLevelDiagLogger(logLevel, logger), true)``` as it continues to hide the complexity.

src/internal/global-utils.ts
```diff
@@ -0,0 +1,98 @@
...
+    GLOBAL_OPENTELEMETRY_API_KEY
+  ] ?? {
+    version: VERSION,
+  };
+
+  const api = _global[GLOBAL_OPENTELEMETRY_API_KEY]!;
+  if (!allowOverride && api[type]) {
+    // already registered an API of this type
+    diag.error(
+      String(
```
Flarna, line 45: Isn't this just a complicated variant of prepending `Error: `?
If you want the call stack you should use `new Error('...').stack`.

It seems also that this path is not covered by a test. I wonder where this log is sent to when it happens during registration of diag.

Flarna:
Maybe create a separate PR for the diag refactoring? At least for me it's hard to follow this PR here which says it's about semver checks but more or less all discussions and a large amount of the changes are about diagnostics.

Please update PR description in first message at some allow people to get some hints what this is good for without the need to read through the old PR.

dyladan:
> Maybe create a separate PR for the diag refactoring? At least for me it's hard to follow this PR here which says it's about semver checks but more or less all discussions and a large amount of the changes are about diagnostics.
> 
> Please update PR description in first message at some allow people to get some hints what this is good for without the need to read through the old PR.

Sorry, the diag refactoring was a requirement for this. I'll split it into another PR to make this one more directed.

src/internal/global-utils.ts
dyladan: Ah yes I needed stack thanks.

>It seems also that this path is not covered by a test. I wonder where this log is sent to when it happens during registration of diag.

I will add a test. If this happens during the registration of diag, it will be sent to whichever diag was already registered. We could send to _both_ diag instances which may in some cases result in duplicate logs but in other cases might prevent the log from being swallowed by incorrect diag use.

test/diag/logLevel.test.ts
dyladan: Since the log level is no longer stored as a separate property but is just enclosed in the registered filtered logger, this drops the level as well.

dyladan:
Closing this PR in order to have the diag refactor in its own PR. Creating a new PR which depends on that PR. I will keep this branch without deleting until I'm sure we don't need it.

Diag refactor: #9 
New global semver PR: #10

PR #9, opened 2021-02-24
chore: refactor diag logger

dyladan:
/cc @MSNev 

Per request of @Flarna this is the diag refactoring split out of #6. There may be some relevant discussion in the previous PR.

This PR is a prerequisite for #10 

- The filtered logger is now the only global that is saved
	- log level and the logger are now enclosed in the filtered logger
- `setLogLevel` is now removed in favor of a second optional argument to `setLogger`
- `setLogger()`/`setLogger(undefined)`/`setLogger(null)` are now disallowed and replaced with `disable`
	- Calling this way anyway results in a noop logger
- Constructing a log level filtered logger results in a logger which ignores the global log level and only uses the one it is specifically constructed with
- Methods to create log level loggers and noop loggers are now not exported to the user and treated as internal concerns. The only exports are `diag`, `DiagLogger`, `DiagLogLevel`, and `LogFunction`.


dyladan:
@MSNev I can't add you as a reviewer because it seems you aren't a member of the OpenTelemetry community?

src/diag/logger.ts
```diff
@@ -59,6 +59,18 @@ export interface DiagLogger {
   verbose: DiagLogFunction;
 }
 
+/**
+ * A filtered diag logger has the same functions as a diag logger, but calls the
+ * child logging methods or not depending on some filtering scheme; for example
+ * log level, namespace, or throttling.
+ */
+export interface FilteredDiagLogger extends DiagLogger {
```
MSNev, line 67: Same comments as previous PR
- Do we need this? Just add this function as optional to the DiagLogger?
- If keep, suggested name of ParentDiagLogger.

dyladan: Adding it to the `DiagLogger` interface adds it to the `diagLoggerFunctions` type too, because that uses `keyof`. I think having the separate type conveys intent well anyway. I will make the name change.

dyladan: Actually the more I think about this, the more I think you're right. There's no reason to expose a filtered logger interface or a parent logger interface. We're just leaking out internals to the user. Being able to get the child of a logger is only useful in the case where we want to fall back if the user uses `undefined`, `null`, or `diag` as the logger. If we simply disallow these cases and fall back to noop, we remove the need for this function entirely.

Further, exporting the constructing functions for noop and log level loggers isn't necessary and just adds API surface for us to maintain into the future. I would prefer to not export them to the user and treat them as internal concepts.

In my mind, we only need to export the following to the user:

```typescript
type LogFunction = (message: string, ...args: unknown[]) => void;

interface DiagLogger {
	verbose: LogFunction;
	debug: LogFunction;
	info: LogFunction;
	warn: LogFunction;
	error: LogFunction;
}

enum DiagLogLevel {
	NONE = 0,
	ERROR = 30,
	WARN = 50,
	INFO = 60,
	DEBUG = 70,
	VERBOSE = 80,
	ALL = 9999,
}

class DiagAPI implements DiagLogger {
	// diag logger functions

	setLogger(logger: DiagLogger, logLevel = DiagLogLevel.INFO): void;
	disable(): void;
}
```

Everything else should be considered and API internal concern

MSNev: Agree, we might get some push back (issues) created from existing users that where using ```NoopLogger()``` and now (0.17) using ```createLogLevelDiagLogger```, but I think that is would be better to take the pain and only if necessary add it back 😃, as really with the refactoring they should not be using it directly anyway...

dyladan: Take a look at this badceec33ba3e3e42851fb150614d05f142c2bd5

dyladan:
> Ok, I know I said id mention that "m" word less, buuuutttt...
> 
> Does this result in smaller output? Looks like it might be close -- enough?
> 
> As this definitely looks easier to read. And while I can think of a few additional tricks, I'm not sure its worth it, basically
> 
> * keep the `diagLoggerFunctions` but don't export from main project (effectively keeping internal)
> * still use for loop for `diag.ts`, here (which some funcky toUpper() for the level lookup) and in the noopLoger.ts.

The looping might trim a tiny bit off, but you also add the array itself and the import statements.

vmarchaud (approved):

src/api/diag.ts
```diff
@@ -71,14 +66,13 @@ export class DiagAPI implements DiagLogger {
...
-    let _filteredLogger: DiagLogger | null;
-    let _logger: DiagLogger = createNoopDiagLogger();
+    const _noopLogger = createNoopDiagLogger();
+    let _filteredLogger: DiagLogger | undefined;
 
     function _logProxy(funcName: keyof DiagLogger): DiagLogFunction {
       return function () {
         const orgArguments = arguments as unknown;
-        const theLogger = _filteredLogger || _logger;
+        const theLogger = _filteredLogger || _noopLogger;
```
obecny, line 75: This check (and next in line 83/77) is done with every call to _logProxy which could be optimised so that when a logger is set all these "decisions" are made once only and it will either use noop or the correct function. It will produce unnecessary overhead.

src/api/diag.ts
```diff
@@ -94,57 +88,57 @@ export class DiagAPI implements DiagLogger {
...
 
-    self.setLogLevel = (maxLogLevel: DiagLogLevel) => {
-      if (maxLogLevel !== _logLevel) {
-        _logLevel = maxLogLevel;
-        if (_logger) {
-          _filteredLogger = createLogLevelDiagLogger(maxLogLevel, _logger);
-        }
-      }
+    self.disable = () => {
+      _filteredLogger = undefined;
```
obecny, line 113: this is set to undefined and then it is checked every time the log function is being called, it should be set here to noop already.

src/diag/internal/logLevelLogger.ts
```diff
@@ -0,0 +1,62 @@
...
+  function _filterFunc(
+    theLogger: DiagLogger,
+    funcName: keyof DiagLogger,
+    theLevel: DiagLogLevel
+  ): DiagLogFunction {
+    if (maxLevel >= theLevel) {
+      return function () {
+        const orgArguments = arguments as unknown;
+        const theFunc = theLogger[funcName];
+        if (theFunc && typeof theFunc === 'function') {
```
obecny, line 44: it is the same pattern here, those checks should be done once only during the creation of log level diag logger, it is an extra cpu time that is not needed - it should either map function or noop without any more "if"

src/api/diag.ts
dyladan: I will add a shortcut to the noop proxy function which avoids all checks if the logger is undefined.

src/api/diag.ts
dyladan: These log proxy functions are on the DiagAPI instance itself. In order to remove this check safely we would have to patch our own instance each time the logger is changed. This is small overhead for a significant safety gain

src/diag/internal/logLevelLogger.ts
dyladan: i'll look into doing these checks at construction

src/api/diag.ts
dyladan: Thinking more I think you're right these checks are redundant. The only thing that can be in _filteredLogger is the logger created by us which is guaranteed to be correct. Only way to break it is to monkey patch things in the API and we can't stop users from hurting themselves.

obecny: 1.
```
  let _filteredLogger: DiagLogger = createNoopDiagLogger();
```
2. 
```
    function _logProxy(funcName: keyof DiagLogger): DiagLogFunction {
      return function () {
        const orgArguments = arguments as unknown;
        const theFunc = _filteredLogger[funcName];
        return theFunc.apply(
          _filteredLogger,
          orgArguments as Parameters<DiagLogFunction>
        );
      };
    }

    for (let i = 0; i < diagLoggerFunctions.length; i++) {
      const name = diagLoggerFunctions[i];
      self[name] = _logProxy(name);
    }

```
3. And createLogLevelDiagLogger
```
  const newLogger = {} as DiagLogger;
  const noop = function () {};
  for (let i = 0; i < levelMap.length; i++) {
    const name = levelMap[i].n;
    const f = levelMap[i].l > maxLevel ? noop : logger[name] || noop;
    newLogger[name] = f;
  }


```

_filteredLogger is always either a noop logger or the one created by `createLogLevelDiagLogger`
no need to check for function existence anymore as this is done on creation

obecny (approved):
lgtm

src/diag/internal/logLevelLogger.ts
```diff
@@ -0,0 +1,51 @@
...
+  logger: DiagLogger
+): DiagLogger {
+  if (maxLevel < DiagLogLevel.NONE) {
+    maxLevel = DiagLogLevel.NONE;
+  } else if (maxLevel > DiagLogLevel.ALL) {
+    maxLevel = DiagLogLevel.ALL;
+  }
+
+  // In case the logger is null or undefined
+  logger = logger ?? {};
```
MSNev, line 30: Using ```||``` is better from a size perspective and as this should always be an object, undefined or null the result is the same.

MSNev (approved):
looks good -- just added 1 suggestion

src/diag/internal/logLevelLogger.ts
dyladan: sure

MSNev:
> @MSNev I can't add you as a reviewer because it seems you aren't a member of the OpenTelemetry community?

https://github.com/open-telemetry/community/issues/666

PR #10, opened 2021-02-24
Use semver to determine API compatibility

dyladan:
Replaces #6 and depends on #9. ~~This will be kept as a draft until #9 merges.~~

This PR introduces the ability for the API to use its version to determine compatibility with the global API instance in a semver compatible manner, replacing the `API_COMPATIBILITY_VERSION` constant which had to be manually incremented.

In order to do this, it introduces `registerGlobal` and `unregisterGlobal` functions.

### `registerGlobal`

Takes a string which identifies the API type and an instance of that API. Checks that no API of that name has been registered. Also checks that it is compatible with previously registered API types so that we can be sure interoperation between APIs is safe. This is not exported and is not meant to be called by end users directly.

### `unregisterGlobal`

This is called when the user disables an API. It removes the globally registered API instance identified by the string. This is not exported and is not meant to be called by end users directly.


src/internal/global-utils.ts
```diff
@@ -0,0 +1,92 @@
...
+  if (!allowOverride && api[type]) {
+    // already registered an API of this type
+    const err = new Error(
+      `@opentelemetry/api: Attempted duplicate registration of API: ${type}`
+    );
+    diag.error(err.stack || err.message);
+    return;
+  }
+
+  if (api.version != VERSION) {
```
vmarchaud, line 51: shouldn't this be a strict equal ?

vmarchaud (approved):
nit pick, otherwise lgtm

src/internal/global-utils.ts
Flarna: see #12

Flarna (approved):

obecny (approved):
lgtm

PR #14, opened 2021-02-27
feat: add tracer.startActiveSpan() helper

naseemkullah:
fixes: open-telemetry/opentelemetry-js#1923

Continuing on @Flarna's work from https://github.com/open-telemetry/opentelemetry-js-api/pull/7

Though it seems like https://github.com/open-telemetry/opentelemetry-js/issues/1923 is where namespaced vs non-namespaced decision will be made. I will post there and ask for a tl;dr as to which it should be.

### DIFF
--- a/src/context/context.ts
+++ b/src/context/context.ts
@@ -15,7 +15,7 @@
  */
 
 import { Context } from './types';
-import { Baggage, Span, SpanContext } from '../';
+import { Baggage, Span, SpanContext, context } from '../';
 import { NoopSpan } from '../trace/NoopSpan';
 
 /**
@@ -53,6 +53,28 @@ export function getSpan(context: Context): Span | undefined {
  */
 export function setSpan(context: Context, span: Span): Context {
   return context.setValue(SPAN_KEY, span);
+}
+
+/**
+ * Helper to execute the given function on a new context created from
+ * active context with given span set.
+ * Note: The global context manager is used to get active context.
+ * @param span span to set on context
+ * @param fn function to execute in new context
+ * @param thisArg optional receiver to be used for calling fn
+ * @param args optional arguments forwarded to fn
+ */
+export function withSpan<
+  A extends unknown[],
+  F extends (...args: A) => ReturnType<F>
+>(
+  span: Span,
+  fn: F,
+  thisArg?: ThisParameterType<F>,
+  ...args: A
+): ReturnType<F> {
+  const ctx = setSpan(context.active(), span);
+  return context.with(ctx, fn, thisArg, ...args);
 }
 
 /**
--- a/test/context/context.test.ts
+++ b/test/context/context.test.ts
@@ -17,15 +17,43 @@
 import * as assert from 'assert';
 import {
   createContextKey,
+  getSpan,
   isInstrumentationSuppressed,
   ROOT_CONTEXT,
   suppressInstrumentation,
   unsuppressInstrumentation,
+  withSpan,
 } from '../../src/context/context';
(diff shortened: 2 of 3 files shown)


src/context/context.ts
```diff
@@ -55,6 +55,28 @@ export function setSpan(context: Context, span: Span): Context {
...
+ * active context with given span set.
+ * Note: The global context manager is used to get active context.
+ * @param span span to set on context
+ * @param fn function to execute in new context
+ * @param thisArg optional receiver to be used for calling fn
+ * @param args optional arguments forwarded to fn
+ */
+export function withSpan<
+  A extends unknown[],
+  F extends (...args: A) => ReturnType<F>
```
aabmass, line 69: It would be great if `F` was also passed the span instance so you could write code like
```ts
withSpan(tracer.startSpan('name'), async (span: Span) => {
  // ...
});
```

and avoid having to assign the span to a variable first like:
```ts
let span = tracer.startSpan('foo');
withSpan(span, () => {...});

span = tracer.startSpan('bar');
withSpan(span, () => {...});
```

Flarna: The main idea of putting a span into a context and set this context active is to avoid the need to pass spans around manually. Instead you fetch it via `getSpan(context.active())`.

aabmass: To clarify, I don't mean so it's available to pass around to other functions. I just mean as a convenience in the "with" scope to avoid boilerplate like `getSpan(context.active()).setAttribute('foo', 'bar')` everywhere.
```

Comment (dyladan) [reply src/context/context.ts:69]: The recommended solution for that would be

```typescript
withSpan(tracer.startSpan('name', { attributes }), async () => {
  // ...
});
```

Flarna: Isn't a closure scope exactly providing what you want?
```js
const span = tracer.startSpan('name');
withSpan(span, async () => {
  // ... span is accessible here
});
```

The current `with()` proposal allows to use any existing function without modification as you can pass `thisArg` and arguments as needed. Having a requirement to accept a span as first argument would be quite limiting or requires everyone to implement wrappers/adaptors.

aabmass: @dyladan right, but often you want to set attributes and span status, events, etc. conditionally and then always finally `span.end()`.

I'm not claiming it's not possible to do as is, but this PR is intended to be a convenience API so why not make it even more convenient? It also avoids having to name a bunch of variables when dealing with concurrent operations:

```ts
const span1 = tracer.startSpan('part1');
const span2 = tracer.startSpan('part2');
await Promise.all(
  withSpan(span1, async () => {
    // ...
    span1.end();
  }),
  withSpan(span2, async () => {
    // ...
    span2.end();
  })
);
```
vs
```ts
await Promise.all(
  withSpan(tracer.startSpan('part1'), async (span: Span) => {
    // ...
    span.end();
  }),
  withSpan(tracer.startSpan('part2'), async (span: Span) => {
    // ...
    span.end();
  })
);
```

dyladan:
Based on the spec in `Span Creation` https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span-creation

>In languages with implicit Context propagation, Span creation MUST NOT set the newly created Span as the active Span in the current Context by default, but this functionality MAY be offered additionally as a separate operation.

I think we should have `tracer.startActiveSpan()` which takes all the same arguments as `startSpan`, but also takes as a final argument a callback which is called in the context of the span and receives the newly created span as an argument.

naseemkullah:
closing and will open a new since suggested approach differs from original work and I cannot figure out `This commit does not belong to any branch on this repository.` right now

PR #16, opened 2021-03-03
fix: Reverse the direction of the semver check

dyladan:
The semver check is backwards. It has not yet caused a bug because we haven't released any other 0.18.x

In the current 0.18 semver check if the user registers a global with version 0.18.0, then API module 0.18.1 attempts to access it, it will succeed. 0.18.1 may attempt to call a method which doesn't exist on 0.18.0. It should be the other way around. 0.18.0 should be able to access the 0.18.1 global because that global has at least the methods 0.18.0 expects.

I caught this while testing the semver check with fake versioning in my sample project

### DIFF
--- a/src/internal/semver.ts
+++ b/src/internal/semver.ts
@@ -23,21 +23,27 @@ const re = /^(\d+)\.(\d+)\.(\d+)(?:-(.*))?$/;
  *
  * The returned function has the following semantics:
  * - Exact match is always compatible
- * - Major versions must always match
- * - The minor version of the API module requesting access to the global API must be greater or equal to the minor version of this API
+ * - Major versions must match exactly
+ *    - 1.x package cannot use global 2.x package
+ *    - 2.x package cannot use global 1.x package
+ * - The minor version of the API module requesting access to the global API must be less than or equal to the minor version of this API
+ *    - 1.3 package may use 1.4 global because the later global contains all functions 1.3 expects
+ *    - 1.4 package may NOT use 1.3 global because it may try to call functions which don't exist on 1.3
+ * - If the major version is 0, the minor version is treated as the major and the patch is treated as the minor
  * - Patch and build tag differences are not considered at this time
  *
  * @param ownVersion version which should be checked against
  */
 export function _makeCompatibilityCheck(
   ownVersion: string
-): (version: string) => boolean {
+): (globalVersion: string) => boolean {
   const acceptedVersions = new Set<string>([ownVersion]);
   const rejectedVersions = new Set<string>();
 
   const myVersionMatch = ownVersion.match(re);
   if (!myVersionMatch) {
-    throw new Error('Cannot parse own version');
+    // we cannot guarantee compatibility so we always return noop
+    return () => false;
   }
 
   const ownVersionParsed = {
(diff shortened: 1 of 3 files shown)


vmarchaud (commented):
shouldn't we bump minor when adding new method on the API ? So the issue never happens

dyladan:
Bumping minor when adding a new API method is what we will do.

### Before This Fix

##### app API version behind dependency API version

- End-user application (app) depends on API version 1.3.1
- Dependency of app (dep) depends on API version 1.4.0
- app registers global API 1.3.1
- dep calls `getGlobal` and receives a 1.3.1 compliant object
- dep calls a new method that doesn't exist on that object

##### app API version ahead of dependency API version

- End-user application (app) depends on API version 1.4.0
- Dependency of app (dep) depends on API version 1.3.1
- app registers global API 1.4.0
- dep calls `getGlobal` and receives undefined
- dep uses noop even though it would have been safe to use the global API object

### After This Fix

##### app API version behind dependency API version

- End-user application (app) depends on API version 1.3.1
- New method is added to the API and minor version is incremented
- Dependency of app (dep) depends on API version 1.4.0
- app registers global API 1.3.1
- dep calls `getGlobal` and receives undefined
- dep uses noop so everything is ok

##### app API version ahead of dependency API version

- End-user application (app) depends on API version 1.4.0
- Dependency of app (dep) depends on API version 1.3.1
- app registers global API 1.4.0
- dep calls `getGlobal` and receives a 1.4.0 compliant object
- dep only knows methods that existed on 1.3.1 and still exist on 1.4.0 so everything is ok

src/internal/semver.ts
```diff
@@ -23,21 +23,27 @@ const re = /^(\d+)\.(\d+)\.(\d+)(?:-(.*))?$/;
...
+ *    - 2.x package cannot use global 1.x package
+ * - The minor version of the API module requesting access to the global API must be less than or equal to the minor version of this API
+ *    - 1.3 package may use 1.4 global because the later global contains all functions 1.3 expects
+ *    - 1.4 package may NOT use 1.3 global because it may try to call functions which don't exist on 1.3
+ * - If the major version is 0, the minor version is treated as the major and the patch is treated as the minor
  * - Patch and build tag differences are not considered at this time
  *
  * @param ownVersion version which should be checked against
  */
 export function _makeCompatibilityCheck(
```
vmarchaud, line 37: makes me wonder if we couldn't just vendor `semver` package ? Seems error prone to re-implement the logic ourselves. That would weight for browser but i'm just throwing the idea out there

test/internal/semver.test.ts
```diff
@@ -92,10 +94,11 @@ describe('Version Compatibility', () => {
...
-    it('should be incompatible if minor versions match but other patch version is greater than our patch version', () => {
+    it('should be compatible if minor versions match but other patch version is greater than our patch version', () => {
+      // global is ahead of us, but changes are backwards compatible
       const check = _makeCompatibilityCheck('0.3.3');
-      assert.ok(!check('0.3.4-alpha'));
-      assert.ok(!check('0.3.4'));
+      assert.ok(check('0.3.4-alpha'));
+      assert.ok(check('0.3.4'));
     });
   });
```
vmarchaud, line 103: could we add another test cases that verify that its work if minor are equal under major `0` ? 
```ts
 it('should be compatible if major and minor versions are equal', () => {
      const check = _makeCompatibilityCheck('0.2.3');
      assert.ok(check('0.2.2'));
      assert.ok(check('0.2.2-alpha'));
      assert.ok(check('0.2.4'));
      assert.ok(check('0.2.4-alpha'));
    });

```

vmarchaud (commented):
overall fine, nit pick for the test

src/internal/semver.ts
dyladan: The semver satisfies is more strict than we are. This could be a good thing though. If global is `1.3.3`, we will be able to call it from `1.3.2` and `1.3.4`. With the semver package we would only be able to call it from `1.3.2`. Semver will never say an earlier version of a package satisfies. This would require end-user applications to always have the latest patch versions. Honestly may not be such a bad thing.

test/internal/semver.test.ts
dyladan: When major is 0, minor is treated like major would be normally. This is tested in the `0.x` section.

src/internal/semver.ts
Flarna: Well, you can pimp a version like 1.3.3 to `1.3.x` then `semver.satisfies()` will be fine with 1.3.0 and 1.3.5.
semver offers a lot APIs to parse/modify versions. It's also possible to e.g. include pre-releases.

I think if we need something in version checking which semver can't do it's most likely problematic.

Main question is do we want one more dependency in API. Don't think there is any real world node application out there which hasn't semver already in it's node_modules tree somewhere.
Can't tell for browser.

dyladan: Without semver:
```
test/index-webpack.js    3.49 MiB  test/index-webpack
```

With semver:
```
test/index-webpack.js    3.66 MiB  test/index-webpack
```

Looks like a 0.17 MiB difference if we add the dependency.

dyladan: >Don't think there is any real world node application out there which hasn't semver already in it's node_modules tree somewhere

In my experience this is about right

>Can't tell for browser.

I suspect it's much lower since semver is mostly used for dependency management and the browser doesn't often have reason for that after compilation

dyladan: >Well, you can pimp a version like 1.3.3 to 1.3.x then semver.satisfies() will be fine with 1.3.0 and 1.3.5.
semver offers a lot APIs to parse/modify versions. It's also possible to e.g. include pre-releases.

Yes these things are for sure possible. We need to decide what behavior we want in this regard either way.

src/internal/semver.ts
```diff
@@ -46,66 +52,73 @@ export function _makeCompatibilityCheck(
...
     }
 
-    const otherVersionParsed = {
-      major: +m[1],
-      minor: +m[2],
-      patch: +m[3],
+    const globalVersionParsed = {
+      major: +globalVersionMatch[1],
+      minor: +globalVersionMatch[2],
+      patch: +globalVersionMatch[3],
```
MSNev, line 84: Semver also has "-<pre-release>+<build>", considering we are about to release an "-RC1" this starts to get complicated...
By the time we add the extra checks how much more of that 0.17Mib would get consumed, while I'll big on minification, I'm also big on don't re-invent the wheel (unless absolutely necessary for perf or size). And as most CDN payloads are also GZip'd how much does that 0.17 shrink too?

MSNev: ok, just noticed the tests below using "alpha".. so ignore the above about pre-release and build#

dyladan: Someone note the time @MSNev advocated for an increased bundle size 🤣

src/internal/semver.ts
vmarchaud: > We need to decide what behavior we want in this regard either way.

If we don't get concensus right now, i think we could still refactor this later as its purely internal.

dyladan: I revamped the test suite to be much more comprehensive and much more clear 2b60af951835615482a5658238f603eec0ccce16. Makes me more comfortable leaving the semver package out of it.

vmarchaud (approved):

src/internal/semver.ts
dyladan: > its purely internal.

It's not purely internal. If we adopt semver, we will get stricter checks for prerelease versions. If we release our version, then `1.0.0`, `1.0.0-alpha.1`, and `1.0.0-alpha.2` will all be compatible with each other in both directions, which means we will be unable to make _any_ changes once we release RC.

obecny (approved):

PR #20, opened 2021-03-08
docs: document latest manual tracing

dyladan:
Document manual tracing using the most up-to-date APIs

Placeholders for context, propagation, and library author documentation as well as glossary of terms.

Lint task fails because links refer to files which do not yet exist but are added by this PR


docs/tracing.md
```diff
@@ -0,0 +1,126 @@
...
+This quick start is for end users of OpenTelemetry who wish to manually trace their applications. If you are a library author, please see the [Library Authors Guide](library-author.md). If you wish to automatically instrument your application, see the automatic instrumentation documentation for the SDK you wish to use.
+
+_Trace API reference: <https://open-telemetry.github.io/opentelemetry-js/classes/traceapi.html>_
+
+### Acquiring a Tracer
+
+In OpenTelemetry, tracing operations are performed using methods on a _tracer_. You can get a tracer by calling [`getTracer`](https://open-telemetry.github.io/opentelemetry-js/classes/traceapi.html#gettracer) on the global tracer provider. `getTracer` takes the name and version of the application or library acquiring the tracer, and provides a tracer which can be used to trace operations.
+
+```typescript
+import { trace } from '@opentelemetry/api';
```
Flarna, line 12: I'm fine with using typescript in samples/docs but till now javascript is usually used everywhere.

docs/tracing.md
```diff
@@ -0,0 +1,126 @@
...
+// When a span is ended, it will be exported to a tracing backend
+// via the currently registered SDK.
+span.end();
+```
+
+Most of the time, spans will be used as part of a function which responds to some event like a web request. The following example shows what it might look like to manually trace a function which responds to a get request using an imaginary http server framework.
+
+```typescript
+async function onGet(request, response) {
+  const span = tracer.startSpan("GET");
```
Flarna, line 37: I think it would be good to mention also `SpanKind` somewhere.

docs/tracing.md
```diff
@@ -0,0 +1,126 @@
...
+  span.end();
+
+  // Attributes MAY NOT be added after the span ends
+  span.setAttribute("my.attribute", false); // this does nothing
+}
+```
+
+#### Semantic Conventions
+
+One problem with span names and attributes is recognizing, categorizing, and analyzing them in your tracing backend. Between different applications, libraries, and tracing backends there might be different names and expected values for various attributes. For example, your application may use `http.status` to describe the HTTP status code, but a library you use may use `http.status_code`. In order to solve this problem, OpenTelemetry uses a library of semantic conventions which describe the name and attributes which should be used for specific types of spans.
```
Flarna, line 85: Maybe extend this a little bit. e.g. recommended usage where applicable but mention also that other keys are allowed.

docs/tracing.md
```diff
@@ -0,0 +1,126 @@
...
+
+Now that we can start and end spans and effectively describe them, it is useful to be able to describe their relationships to each other. For instance, if one span describes an incoming request which makes a database call, it may be useful to trace the database call as a separate span which is a child of the original request span. In order to do this, when we create a span we can tell OpenTelemetry which span to use as its parent using a mechanism called _Context_.
+
+Context is a very important part of the OpenTelemetry API which cannot be adequately explained in a single paragraph. To read more about context, see the [context documentation](context.md).
+
+```typescript
+import { context, setSpan } from '@opentelemetry/api';
+
+async function onGet(request, response) {
+  const span = tracer.startSpan("GET /user/:id", {
```
Flarna, line 99: should be most likely 
```typescript
const span = tracer.startSpan("GET /user/:id", {
  attributes: {
    "http.method": "GET",
    "http.url": request.url,
  }
});
```
I think we should use semantic conventions in our samples if applicable - and SpanKind.

docs/tracing.md
```diff
@@ -0,0 +1,126 @@
...
+    "http.url": request.url,
+  });
+
+  const userId = request.params.id;
+
+  // Create a new context from the current context which has the span "active"
+  const ctx = setSpan(context.active(), span);
+  
+  // Call getUser with the newly created context
+  const user = await context.with(ctx, getUser, userId);
```
Flarna, line 110:
```suggestion
  const user = await context.with(ctx, getUser, undefined, userId);
```
The first argument after the function is the receiver.

README.md
```diff
@@ -7,141 +7,70 @@
...
+In order to get started with tracing, you will need to first register an SDK. The SDK you are using may provide a convenience method which calls the registration methods for you, but if you would like to call them directly they are documented here: [sdk registration methods](docs/sdk-registration.md).
 
-- [Trace API Documentation][trace-api-docs]
-- [Propagation API Documentation][propagation-api-docs]
-- [Context API Documentation][context-api-docs]
+Once you have registered an SDK, you can start and end spans. A simple example of basic SDK registration and tracing a simple operation is below. The example should export spans to the console once per second. For more information, see the [tracing documentation](docs/tracing.md).
 
-```javascript
-const api = require("@opentelemetry/api");
+```typescript
```
Flarna, line 34: actually it's javascript here.

README.md
```diff
@@ -7,141 +7,70 @@
...
- *
- * By default the NodeTracerProvider uses Trace Context for propagation
- * and AsyncHooksScopeManager for context management. To learn about
- * customizing this behavior, see API Registration Options below.
- */
-tracerProvider.register();
-```
+- An application you wish to instrument
+- [OpenTelemetry JS SDK][opentelemetry-js]
+- Node.js 8+ (14+ is preferred) or an ECMAScript 5+ compatible browser
```
Flarna, line 18:
```suggestion
- Node.js >=8.5.0 (14+ is preferred) or an ECMAScript 5+ compatible browser
```
Is `ECMAScript 5+` really new enough for the samples here using `await` and `@opentelemetry/tracing` which is transpiled to ES2017 (like the whole SDK).

docs/tracing.md
dyladan: Only difference is `import` vs `require` for now. None of the code examples have types in them

README.md
```diff
@@ -152,6 +81,8 @@ async function doSomething() {
 
 Apache 2.0 - See [LICENSE][license-url] for more information.
 
+[opentelemetry-js]: https://github.com/open-telemetry/opentelemetry-js
```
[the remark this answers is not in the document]
dyladan: I thought I had these fixed but guess there was a stash/pop issue that lost them maybe

README.md
dyladan: I will make it clear that the API itself requires 5+ but the sdk may need newer. I can also update the examples to not use async/await, I just thought it was clearer for the examples.

dyladan:
@Flarna think i've addressed all your comments. Would appreciate another pass if you have time

docs/tracing.md
```diff
@@ -0,0 +1,164 @@
...
+    kind: SpanKind.SERVER
+  });
+
+  const userId = request.params.id;
+
+  // Create a new context from the current context which has the span "active"
+  const ctx = setSpan(context.active(), span);
+  
+  // Call getUser with the newly created context
+  const user = await context.with(ctx, getUser, undefined, userId);
```
vmarchaud, line 85: i think we could inline the callback here, the fact that we are giving its param with the 4th arg is kinda complex to understand for beginner i think

docs/tracing.md
```diff
@@ -0,0 +1,164 @@
...
+}
+
+async function getUser(userId) {
+  // when this span is created, it will automatically use the span from the context as its parent
+  const span = tracer.startSpan("SELECT ShopDb.Users", {
+    attributes: {
+      "db.system": "mysql",
+      "db.connection_string": "Server=shopdb.example.com;Database=ShopDb;Uid=billing_user;TableCache=true;UseCompression=True;MinimumPoolSize=10;MaximumPoolSize=50;",
+      "db.user": "app_user",
+      "net.peer.name": "shopdb.example.com",
```
vmarchaud, line 104: could we refer to semantic conventions here to explain that those are "standard" attributes ?

docs/glossary.md
```diff
@@ -0,0 +1,23 @@
+# Glossary of Terms
```
vmarchaud, line 1: does it make sense to document it there even though its already explained in the spec ?

docs/tracing.md
```diff
@@ -54,73 +54,111 @@ async function onGet(request, response) {
...
+}
 
-_See the current trace semantic conventions in the OpenTelemetry Specification repository: <https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/trace/semantic_conventions>_
+server.on("GET", "/user/:id", onGet);
+```
 
 ### Span Relationships
 
-Now that we can start and end spans and effectively describe them, it is useful to be able to describe their relationships to each other. For instance, if one span describes an incoming request which makes a database call, it may be useful to trace the database call as a separate span which is a child of the original request span. In order to do this, when we create a span we can tell OpenTelemetry which span to use as its parent using a mechanism called _Context_.
+One of the most important aspects of spans is their relationships to each other. For instance, if one span describes an incoming request which makes a database call, it may be useful to trace the database call as a separate span which is a child of the original request span. In order to do this, when we create a span we can tell OpenTelemetry which span to use as its parent using a mechanism called _Context_.
```
Flarna, line 126:
```suggestion
One of the most important aspects of spans is their relationships to each other. For instance, if one span describes an incoming request which makes a database call, it's recommended to trace the database call as a separate span which is a child of the original request span. In order to do this, when we create a span we can tell OpenTelemetry which span to use as its parent using a mechanism called _Context_.
```

docs/tracing.md
```diff
@@ -0,0 +1,164 @@
...
+
+```typescript
+import { trace } from '@opentelemetry/api';
+
+const tracer = trace.getTracer("my-application", "0.1.0");
+```
+
+## Starting and Ending a Span
+
+In OpenTelemetry, all _traces_ are composed of [`Spans`](https://open-telemetry.github.io/opentelemetry-js/interfaces/span.html). These spans are linked together by parent-child relationships to form a tree. The resultant tree is your trace, and the root of the tree is commonly called the _root span_.
```
Flarna, line 19: Maybe add `A span should describe a single operation like a database request, an incoming web request,...`.

docs/tracing.md
dyladan: I think if a user has trouble with this then the tsdoc needs to be made more clear. This is the pattern I expect most users to use, not just advanced ones.

docs/glossary.md
vmarchaud: There is one: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md but not really complete, for missing part i think we can just link what the spec explain no ? (ex for [span](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span))

docs/tracing.md
vmarchaud: Do you mean adding an exemple inside the tsdoc directly ? I'm fine with that, i think this would be much clearer:
```ts
const userId = request.params.id;
// Create a new context from the current context which has the span "active"
const ctx = setSpan(context.active(), span);
// Call getUser with the newly created context
const user = await context.with(ctx, async () => {
  // when this span is created, it will automatically use the span from the context as its parent
  const span = tracer.startSpan("SELECT ShopDb.Users", {
    attributes: {
      // Attributes from the database trace semantic conventions
      // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md
      "db.system": "mysql",
      "db.connection_string": "Server=shopdb.example.com;Database=ShopDb;Uid=billing_user;TableCache=true;UseCompression=True;MinimumPoolSize=10;MaximumPoolSize=50;",
      "db.user": "app_user",
      "net.peer.name": "shopdb.example.com",
      "net.peer.ip": "192.0.2.12",
      "net.peer.port": 3306,
      "net.transport": "IP.TCP",
      "db.name": "ShopDb",
      "db.statement": `Select * from Users WHERE user_id = ${userId}`,
      "db.operation": "SELECT",
      "db.sql.table": "Users",
    },
    kind: SpanKind.CLIENT,
  });
  const user = await db.select("Users", { id: userId });
  span.end();
  return user;
})
```

WDYT ?

dyladan: What if I annotate the call to context.with like this? I can change to the inline style if you still think that's more clear, but I think this calling convention will be the more common and recommended approach most of the time.

```javascript
  // Call getUser with the newly created context
  // 
  // context.with calls a function with an associated "active" context. Within
  // the function, calling context.active() returns the currently active context.
  // If there is no active context, the ROOT_CONTEXT will be returned, which
  // has no key-value pairs.
  // 
  // context.with requires at least 2 arguments: a context and a function to be called.
  // If a third argument is provided, it will be bound to `this` `this` inside the function.
  // Any additional parameters are used as arguments when calling the function.
  // 
  //   Return value is the value returned from getUser
  //    |                         Context to be used as the "active" context
  //    |                         |    Function to be called
  //    |                         |    |        Object assigned to this during function execution
  //    |                         |    |        |          Passed as the first argument to getUser
  //    |                         |    |        |          |
  //    V                         V    V        V          V
  const user = await context.with(ctx, getUser, undefined, userId);
```

vmarchaud: > I think this calling convention will be the more common and recommended approach most of the time.

I doesn't expect end users to understand this but thats just a personal opinion. 

> What if I annotate the call to context.with like this

I'm not strictly against this way so i'll go and approve the PR, we'll see later on if thats clear enough. Since you already wrote the detailed explaination i think you can add it, its better than nothing

vmarchaud (approved):

docs/tracing.md
dyladan: There will be a whole md file dedicated to context as well which will contain multiple different calling conventions and the relative benefits of each.

Issue #27, opened 2021-03-31
Return suppressed instrumenation context from noop context manager


dyladan:
> Is this something we can/should improve in API? The main idea was to return Noop instances if version doesn't match but it should not crash nor end up in endless loops.

I believe that it does return NOOP. Our endless loop happened due to the context not being set and read by the same api version, causing  not to work - which is a recipe for infinite loops.
There might be other issues such as [propagation not working](https://github.com/aspecto-io/opentelemetry-ext-js/issues/88) etc, but it will probably resolve when we move to api v.1.0.0 everywhere.

_Originally posted by @blumamir in https://github.com/open-telemetry/opentelemetry-js/issues/2055#issuecomment-810845529_

We can prevent this endless loop by returning a suppressed instrumentation context from the noop context manager. If the user is manually tracing without a context manager, then calling `context.active()` should not be used anyway.

One important caveat to the above is that a call to `startSpan` will by default grab the active context by default. If this context has instrumentation suppressed then the returned span will be noop. Possibly we could introduce a new constant context similar to ROOT_CONTEXT but which is actually NOOP_CONTEXT (only returned by the noop context manager) which can be used to determine if the context manager is enabled?

dyladan:
One open question: is this considered a breaking change? @open-telemetry/javascript-maintainers

Flarna:
Does this really solve a problem? If there is all noop it's don't care if we return ROOT_CONTEXT or NOOP_CONTEXT as the Tracer in question is a NoopTracer.
If we have a mixmax of API/SDK versions with ContextManager, TracerProvider,.. "partly" present and mixed with Noops it seems more luck that the path choosen is that one solved here.

I think there are by far too much places where `context.active()` is used as default to ensure that this is has no side effects. If we go this path I would prefer to remove all this automatism/defaulting.

Just think about the following (ref [here](https://github.com/open-telemetry/opentelemetry-js/blob/62f8695d3fe5309c62418a1043bc6e8a176bc11f/packages/opentelemetry-instrumentation-grpc/src/grpc/index.ts#L200)): `api.propagation.extract(context.active(), carrier)` - this would result in a suppressing context even if a valid span is extracted.
I know this can be fixed by using `ROOT_CONTEXT` instead but how much code is in the wild using `context.active()`?

So in short if we do this I would consider it as breaking.

[...]

Issue #15, opened 2021-03-01
Release candidate


dyladan:
Is it sufficient to put on the README for the API that 0.18 is the release candidate? Or do we want to release something like 1.0.0-rc.1?

_Originally posted by @dyladan in https://github.com/open-telemetry/opentelemetry-js/issues/1982#issuecomment-788245393_

vmarchaud:
I think we can tag it both on github and npm ?

dyladan:
> I think we can tag it both on github and npm ?

To me it seems like there are a few places we could tag the current version as RC:

- GitHub README
- git tag (`rc1` or similar?)
- npm release tags (`rc` maybe?)

In addition to that, we could release a version like `1.0.0-rc.1` which follows semver conventions for a prerelease.

The question is: is it sufficient to tag 0.18 as our release candidate? Or do we need to release a semver prerelease version also?

vmarchaud:
looking at other SIGS:
- python made a release: https://github.com/open-telemetry/opentelemetry-python/releases/tag/v1.0.0rc1
- java directly released 1.0: https://github.com/open-telemetry/opentelemetry-java/blob/main/CHANGELOG.md#version-100---2021-02-26
- .net did few RCs: https://github.com/open-telemetry/opentelemetry-dotnet/tags
- go hasn't made a RC yet

Do we actually expect to run RC release ? If not i'm not sure making a full release would be that useful. Generally i would go with git tag + npm releases to tag it somewhere.

legendecas:
If we define such release as RC, I'd vote for semver like `1.0.0-rc.1` to explicitly say that this version is ready for a stable release. And people can start using `^1.0.0-rc.1` version range, which would get updated if we released stable `1.0.0`.

dyladan:
This is why I wanted it to be the same as 0.18. If someone installs `npm i @opentelemetry/api` they will get `0.18.0`, but using npm tags we can make it so if they explicitly install `npm i @opentelemetry/api@rc` they can get `1.0.0-rc.1` which will automatically be updated to `1.0.0`. This means we can "safely" add breaking changes to the RC if absolutely required because users have explicitly opted into it.

vmarchaud:
I see, i think adding a note in the readme + a tag on github (that point to the same commit as 0.18.0) for historic purpose would be fine ?

Flarna:
> This means we can "safely" add breaking changes to the RC

[...]

PR #30, opened 2021-04-02
feat: make context.active() the default value for certain context functions

naseemkullah:
> This is actually discussed there https://github.com/open-telemetry/opentelemetry-js-api/issues/27  for another reason though

functions where I'm adding a default context as `context.active()`, based on searching the otel-js repo for their invocations:
- `getSpan()`
- `setSpan()` 
- `suppressInstrumentation()`

💥  breaks `setSpan()` as it reverses the order of params.

This PR also serves a discussion. Is this a better default or not?

Pros: it will reduce boilerplate code, helper functions aside/when working with lowest level api

Cons: ?

### DIFF
--- a/src/context/context.ts
+++ b/src/context/context.ts
@@ -15,7 +15,7 @@
  */
 
 import { Context } from './types';
-import { Baggage, Span, SpanContext } from '../';
+import { Baggage, context, Span, SpanContext } from '../';
 import { NoopSpan } from '../trace/NoopSpan';
 
 /**
@@ -39,20 +39,20 @@ const BAGGAGE_KEY = createContextKey('OpenTelemetry Baggage Key');
 /**
  * Return the span if one exists
  *
- * @param context context to get span from
+ * @param ctx context to get span from
  */
-export function getSpan(context: Context): Span | undefined {
-  return (context.getValue(SPAN_KEY) as Span) || undefined;
+export function getSpan(ctx = context.active()): Span | undefined {
+  return (ctx.getValue(SPAN_KEY) as Span) || undefined;
 }
 
 /**
  * Set the span on a context
  *
- * @param context context to use as parent
- * @param span span to set active
+ * @param span to set
+ * @param ctx context to use as parent
  */
-export function setSpan(context: Context, span: Span): Context {
-  return context.setValue(SPAN_KEY, span);
+export function setSpan(span: Span, ctx = context.active()): Context {
+  return ctx.setValue(SPAN_KEY, span);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/context/context.ts
```diff
@@ -39,20 +39,20 @@ const BAGGAGE_KEY = createContextKey('OpenTelemetry Baggage Key');
...
 /**
  * Set the span on a context
  *
- * @param context context to use as parent
+ * @param ctx context to use as parent
  * @param span span to set active
  */
-export function setSpan(context: Context, span: Span): Context {
-  return context.setValue(SPAN_KEY, span);
+export function setSpan(ctx = context.active(), span: Span): Context {
```
aabmass, line 54: Correct me if I'm wrong, but now to get the default you have to do `setSpan(undefined, span)`. This doesn't seem like much of an improvement. Could we change the parameter order so it can be optional, or is this that too invasive a change at this point?

naseemkullah: good point.

We might as well reverse the order, or provide an opts obj as the parameter. What would you prefer?

naseemkullah: sidenote: it might be more comprehensible to rename this function to `activateSpan` as per the jsdoc it sets a span to `active`

Flarna: I think the doc is outdated. A few month ago there were seperate context keys for active span and extracted span.
This API just sets a span in a context object. The context itself is not set as the current/active context (that is done by `context.with()`).

naseemkullah: Thanks, I've removed the misleading doc and reversed the params, pardon the force push as i wanted to rename the commit msg with a `!` to indicate breakage. If this is approved I'll add relevant upgrade notes

dyladan:
>This PR also serves a discussion. Is this a better default or not?

In the past we have been moving in the opposite direction. For low-level API primitives we have typically considered it to be better to have more explicit parameters.

/cc @Flarna because I know you've expressed opinions in the past that you believe we already have too many cases where context has a default value of `active()`

Flarna:
I removed the `context.active()` as default for `propagation.extract()` and `propagation.inject()` in https://github.com/open-telemetry/opentelemetry-js/pull/1734
As mentioned there at least in `extract` using `context.active()` (which may contain a span reference) is most likely what noone wants.

The helper functions here are somewhat different as they are used more frequently at places where `context.active()` is what someone wants - similare as in `tracer.startSpan()`.

I'm not a big fan of using references to globals in default arguments as it makes it terrible hard to consistent use non global `ContexManger`, `TracerProvider`,...

I'm also not sure if a breaking change is still acceptable in current stage (see https://github.com/open-telemetry/opentelemetry-js-api/issues/15#issuecomment-797537651).

naseemkullah:
Thanks for the detailed explanation! Closing

PR #32, opened 2021-04-07
feat!: return boolean success value from setGlobalXXX methods

### LINKED ISSUE #31: Return value of setGlobalXXX doesn't match documentation
According to the documentation `api.context.setGlobalContextManager()` _Returns the initialized context manager_.

But actually it returns the `ContextManager` passed even if `registerGlobal()` detects that there is already a context manager installed and the passed `ContextManager` is not installed as global.

Similar `api.propagation.setGlobalPropagator()` returns the propagator passed instead that one installed global.

For `api.trace.setGlobalTracerProvider()` there is even a difference if installation happens as it always returns `this._proxyTracerProvider`.

I'm not sure what's correct here.
* Returning the current installed global version will cause problems in case of version mismatch
* Returning the passed in version may result in having two ContextManagers/TracerProviders/Propagators flying around

I think the best would be to return what one would get if they call `api.trace.getTracerProvider()` or `api.context._getContextManager()` (currently private) or `api.propagation._getGlobalPropagator()` (currently private). So either the global one (version matches) or Noop.

dyladan:
Fixes #31

### DIFF
--- a/src/api/context.ts
+++ b/src/api/context.ts
@@ -44,13 +44,12 @@ export class ContextAPI {
   }
 
   /**
-   * Set the current context manager. Returns the initialized context manager
+   * Set the current context manager.
+   *
+   * @returns true if the context manager was successfully registered, else false
    */
-  public setGlobalContextManager(
-    contextManager: ContextManager
-  ): ContextManager {
-    registerGlobal(API_NAME, contextManager);
-    return contextManager;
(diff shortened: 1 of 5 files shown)


Flarna (approved):

naseemkullah (approved):

vmarchaud (approved):

obecny (approved):
it would be a breaking change :/ although I think this is the last chance we can really do it.

PR #43, opened 2021-04-22
chore: renaming noop span to non recording span

obecny:
Fixes https://github.com/open-telemetry/opentelemetry-js-api/issues/41

PR #46, opened 2021-04-23
chore: do not export singletons

### LINKED ISSUE #38: Consider not exposing type + singleton
A few components have both their type + singleton exported, e.g. `NOOP_TEXT_MAP_PROPAGATOR` and `NoopTextMapPropagator` (likewise for `NoopTracer`, `NoopTracerProvider`, etc). Unless there's a technical reason to do this on purpose, I'd suggest exposing one or the another, in order to prevent confusion.

dyladan:
Fixes #38


src/api/context.ts
```diff
@@ -87,7 +86,7 @@ export class ContextAPI {
   }
 
   private _getContextManager(): ContextManager {
-    return getGlobal(API_NAME) || NOOP_CONTEXT_MANAGER;
+    return getGlobal(API_NAME) || new NoopContextManager();
```
Flarna, line 89: I think it's quite strange to return a new ContextManager on every invocation. I think not exporting the NOOP_CONTEXT_MANAGER doesn't imply that we have to allocate a new one on every call.

src/trace/NoopTracer.ts
```diff
@@ -55,5 +55,3 @@ function isSpanContext(spanContext: any): spanContext is SpanContext {
     typeof spanContext['traceFlags'] === 'number'
   );
 }
-
-export const NOOP_TRACER = new NoopTracer();
```
Flarna, line 59: The `NOOP_TRACER` is currently used in in Tracer.ts of core to create a `NoopSpan` which is currently not exported (as span creation is only allowed via tracer according to spec).
I think we need at some API to create a suppressed/non sampled span if we remove this one.
Maybe we should keep `NOOP_TRACER` and don't export `NoopTracer`?

src/api/context.ts
vmarchaud: Yeah i agree, the singleton wasn't exported here so i'm not sure why remove it

dyladan: Will update

src/trace/NoopTracer.ts
dyladan: `NOOP_TRACER` returns `NonRecordingSpan`s which I already think is confusing. I would rather have a strong API for getting non recording spans from the tracer API when needed instead of using a somewhat hidden noop tracer singleton. I would maybe rather export the NonRecordingSpan or maybe even reimplement it in the `tracing` package.

Flarna: That sounds like we should rename `NoopTracer` to `NoRecordingTracer` - or don't export it at all.

I vote for exposing something in API for this. [Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.2.0/specification/trace/api.md#span-creation) seems to be quite clear that span creation must be done via a Tracer.

`NoopTracer#startSpan` is not complicated but it holds some logic. I don't think we should dup it into SDKs.

dyladan: @Oberon00 suggested we use https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#wrapping-a-spancontext-in-a-span to wrap an invalid context in a non recording span. This uses only pre-existing APIs and does not require that we export any no-op classes or singletons.

Flarna: Fine for me. The logic to reuse parent SpanContext like in NoopTracer is moved into the hands of the caller. But I think this is acceptable.

dyladan: #50 PR incoming

dyladan: oops mean #49

obecny: > That sounds like we should rename `NoopTracer` to `NoRecordingTracer` - or don't export it at all.
> 
> I vote for exposing something in API for this. [Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.2.0/specification/trace/api.md#span-creation) seems to be quite clear that span creation must be done via a Tracer.
> 
> `NoopTracer#startSpan` is not complicated but it holds some logic. I don't think we should dup it into SDKs.

https://github.com/open-telemetry/opentelemetry-js-api/pull/43

dyladan: In the spec for `NonRecordingSpan` they have:

>If a new type is required for supporting this operation, it SHOULD NOT be exposed publicly if possible (e.g. by only exposing a function that returns something with the Span interface type). If a new type is required to be publicly exposed, it SHOULD be named NonRecordingSpan.

IMO this means we need to not export this class and instead users should do `api.trace.wrapSpanContext` in order to get one

Flarna:
I know that the headline tells do not export singletons. But above I read `... does not require that we export any no-op classes or singletons.`

Currently `NoopTextMapPropagator`, `NoopTracer`,.. are still exported. Shoudl we remove them also?

Flarna (approved):
I guess the followup RP in core will be funny :o).

vmarchaud (approved):

src/trace/NoopTracer.ts
dyladan: @obecny is that explanation good enough? I don't want to mark this converstation as resolved until you're happy

obecny (commented):
it looks fine, but before this one is merged, we should first update core and contrib to avoid problems with being unable to make a release of those packages for a long time. Merging this PR should be a last thing in this task

src/trace/NoopTracer.ts
obecny: hmm I was just pointing the PR for renaming the noopspan here, nothing more

obecny (approved):

PR #47, opened 2021-04-25
chore: move span method for context in trace API #40

### LINKED ISSUE #40: Consider moving setSpan()/getSpan() to its related subsection
That is, that it is accessed through `opentelemetry.trace` instead of `opentelemetry` - which is followed by other SIGs, as the idea is overall having each concern have exclusive logical access to its own key/slot. Likewise for the Baggage portion.

Not a strong feeling though, as things are slightly differently organized for JS. For your consideration.

vmarchaud:
- I moved both `[sg]etSpan`/`[sg]etSpanContext` utils into the trace API
- I also moved `[sg]etBaggage` to the baggage folder (next to `createBaggage`)
	- As discussed in the SIG, we should think about the place for baggage utils, i think it makes more sense in the propagation API 

Fixes #40


Flarna (approved):
As this is a breaking change an entry in upgrade guideline (introduced in #43) should be added

obecny (approved):
lgtm

dyladan (approved):
Once conflicts are resolved this can merge

PR #51, opened 2021-04-27
feat: add function to wrap SpanContext in NonRecordingSpan #49

### LINKED ISSUE #49: [spec] add `trace#wrapSpanContext`
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#wrapping-a-spancontext-in-a-span

Brought up in the spec call

### LINKED ISSUE #50: [spec] Do not export NOOP implementations
The NOOP implementations are currently exported. Particularly the NOOP tracer is used to create NOOP spans. After #51, this will no longer be required. Because the specification does not say to export NOOP implementations, we should remove their exports before 1.0

dyladan:
Fixes #49

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -48,7 +48,6 @@ export {
   isSpanContextValid,
   isValidTraceId,
   isValidSpanId,
-  wrapSpanContext,
 } from './trace/spancontext-utils';
 
 export * from './context/context';
--- a/src/index.ts
+++ b/src/index.ts
@@ -48,6 +48,7 @@ export {
   isSpanContextValid,
   isValidTraceId,
   isValidSpanId,
+  wrapSpanContext,
 } from './trace/spancontext-utils';
 
 export * from './context/context';
--- a/src/api/trace.ts
+++ b/src/api/trace.ts
@@ -14,15 +14,18 @@
  * limitations under the License.
  */
 
-import { ProxyTracerProvider } from '../trace/ProxyTracerProvider';
-import { Tracer } from '../trace/tracer';
-import { TracerProvider } from '../trace/tracer_provider';
-import { isSpanContextValid } from '../trace/spancontext-utils';
 import {
   getGlobal,
   registerGlobal,
   unregisterGlobal,
 } from '../internal/global-utils';
+import { ProxyTracerProvider } from '../trace/ProxyTracerProvider';
+import {
+  isSpanContextValid,
+  wrapSpanContext,
+} from '../trace/spancontext-utils';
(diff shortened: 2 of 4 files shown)


src/index.ts
```diff
@@ -48,6 +48,7 @@ export {
   isSpanContextValid,
   isValidTraceId,
   isValidSpanId,
+  wrapSpanContext,
```
Flarna, line 51: Is it intended to export it addtionally flat on api?

Flarna (commented):
We could use the new API also in `setSpanContext`

src/index.ts
dyladan: This is to make minification obsessed people like @MSNev happy. we discussed it in the SIG recently and it was generally agreed that methods exported on a namespace like `api.trace.wrapSpanContext` are more obvious and discoverable to the user, but direct functions are friendlier to minifiers. As a result we agreed to write the implementation as a function and export it both ways as seen here.

Do you have an argument why we shouldn't do this?

Flarna: well, it's kind of strange to export it at two locations.
I'm not a minification expert but I would guess it can remove/minify that one on top level but that one on `api.trace` will be still there. Seems more a loose - loose situation.

Besides that it's inconsistent to move e.g. `getSpan()`, `getSpanContext()`,.. from top level to trace (see #47) without keeping them on top level.

MSNev: This is the correct way to do it for minification purposes, as the api.trace is just a reference to the exported function, so this is the smallest possible way to do it. As if the exported function is not referenced (directly) then it will be dropped from the exported names but will still be exposed via the api.trace.

And @Flarna is also correct, any reference to anything in the api.trace namespace will also mean that everything exposed via api.trace will never get dropped. So ideally (internally) the SDK / API should avoid using any namespace references as this causes nothing to get dropped.

So this approach really only works if EVERYTHING is also exposed as a function so that this is a feasible approach, this doesn't mean that you have to do everything at once but it does mean that until you drop all internal references the complete minification advantages are not achieved.

Flarna: Reading https://github.com/open-telemetry/opentelemetry-js-api/issues/40#issuecomment-824195038 I would guess that only the namespaced version will survive.

dyladan: Ah yes I had forgotten that converstation. I'll update the PR to match. Sorry @MSNev :/

Flarna (approved):

obecny (approved):
lgtm

vmarchaud (approved):

Issue #52, opened 2021-04-29
Diag logger has no context while using OT in large system


Asafb26:
Diag logger has a really crucial part while debugging a host - for tracing issues or even a non-related logic issue.
Currently, the logs dumped by the diag logger have no context, we cannot understand if the log arrived from OT or anywhere else.
I guess that we need to label it somehow - a by static label `[OpenTelemetry]` or custom label for the client choice.

vmarchaud:
For context this was discussed in https://github.com/open-telemetry/opentelemetry-js/issues/2104

dyladan:
per-package namespaces are great but there might be value in a host prefix too.

PR #53, opened 2021-04-29
feat: add prefix option to DiagConsoleLogger

Asafb26:
see https://github.com/open-telemetry/opentelemetry-js-api/issues/52
Introduced an optional solution for this issue.

### DIFF
--- a/src/diag/consoleLogger.ts
+++ b/src/diag/consoleLogger.ts
@@ -30,10 +30,9 @@ const consoleMap: { n: keyof DiagLogger; c: keyof Console }[] = [
  * {@link createLogLevelDiagLogger}
  */
 export class DiagConsoleLogger implements DiagLogger {
-  constructor() {
+  constructor(prefix?: string) {
     function _consoleFunc(funcName: keyof Console): DiagLogFunction {
-      return function () {
-        const orgArguments = arguments;
+      return function (message: string, ...args: unknown[]) {
         if (console) {
           // Some environments only expose the console when the F12 developer console is open
           let theFunc = console[funcName];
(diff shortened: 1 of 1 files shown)


dyladan:
This will give the same prefix to every log. Multiple tracing components in the same process will still not be able to be differentiated.

Asafb26:
> This will give the same prefix to every log. Multiple tracing components in the same process will still not be able to be differentiated.

yep, this prefix intended to give a context for the host itself, not for OT.
currently, the logs are like that:
```
~/Workspace/test_ot_js > PORT=1301 NAME=my-server npm start
> ts-node app.ts

Applying patch for http@12.18.0
Applying patch for https@12.18.0
[my-server] Listening for requests on http://localhost:1301
http instrumentation incomingRequest http
http instrumentation outgoingRequest http
http.ClientRequest return request
http instrumentation incomingRequest HTTP
[my-server] Got body: {test: 1}
outgoingRequest on response()
outgoingRequest on end()
```
IMO we should have the option to add a prefix for OT logs so the clients debugging a flow will be able to understand the source of each log

Asafb26:
Just saw the reference https://github.com/open-telemetry/opentelemetry-js/issues/2104
probably a better approach, so we have the option to understand which package exactly triggers this log instead of just "OpenTelemetry"

dyladan:
A global prefix could be combined with a namespace for additional information. There is no reason they can't coexist. If you write a custom `Logger` that dumps to a logging library like `pino` or `bunyan` you might get host information "for free" though.

vmarchaud:
I think this has been implemented with https://github.com/open-telemetry/opentelemetry-js-api/pull/62/files so we should close this ?

Asafb26:
reopen this one because the newly `ComponentLogger` is only for components (per-package) namespaces, IE it assumes that we have a global logger already. 
In this case, we want a global prefix for the host itself, so we will have a log context for entire OpenTelemetry logging namespace, WDYT?

vmarchaud:
> reopen this one because the newly ComponentLogger is only for components (per-package) namespaces, IE it assumes that we have a global logger already.
In this case, we want a global prefix for the host itself, so we will have a log context for entire OpenTelemetry logging namespace, WDYT?

I think you could just construct a custom logger that does this and set it as the diag logger. This is why the logger is settable so we don't have to implement any options in the API itself

Issue #23, opened 2021-03-17
Should NoopContextManager.active() return the context set by context.with()


Flarna:
Currently `NoopContextManager.active()` always returns `ROOT_CONTEXT`.

This is somewhat strange if you consider following code as it looks like a basic contract is broken:
```js
api.context.with(ctx, () => {
  api.context.active() // returns ROOT_CONTEXT
});
```

I think even the `NoopContextManager` should at least fulfill the basic contract that `context.active()` within the `with()` callback returns the context set.

In special if I think about [Behavior of the API in the absence of an installed SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.1.0/specification/trace/api.md#behavior-of-the-api-in-the-absence-of-an-installed-sdk) a typical piece of code could be something like this (only a propagator installed but no SDK):
```js
const ctx = api.propagation.extract(ROOT_CONTEXT, carrier);
api.context.with(ctx, () => {
  ...
  api.propagation.inject(api.context.active(), carrier);
});
```

Or is this the intended setup and a user has to create/install a working context manager in addition to the propagator in such setups?

vmarchaud:
I'm not sure, its quite tricky to handle because context loss can happen within `with` callback (simply with async/await) so is it worth to change this behavior ?
I think that would be better to force the user to use a working context manager, that would avoid problems where the context api "kinda" work depending on what code is in the callback.

Flarna:
I'm also unsure. If we change it we have to at least enable it [here](https://github.com/open-telemetry/opentelemetry-js-api/blob/504041a8aa8442987a95f675af0a402084678682/src/api/context.ts#L26).

I did a fast look into Java and Python and it seems they instantiate and use a working ContextManager even in API only case. But I assume it's easier for them as they don't have to support different runtimes (Browser/NodeJs).

obecny:
I think it is fine the way it works now. If we start doing some partial solution in api we might confuse user even more, (why it works for sync but not for async etc.). We should just explain that until the context manager is registered the default behaviour is just noop. What we could do instead is probably show some warning that noop is used, and if you want to change it please register context manager first.

aabmass:
I agree with @Flarna, this is quite dangerous for libraries that propagate context even without the SDK. I guess such libraries should really use explicit context passing for anything propagation related and only use `context.with()` before calling user code, but its not obvious right now.

> What we could do instead is probably show some warning that noop is used, and if you want to change it please register context manager first.

[...]

PR #54, opened 2021-05-01
feat: add tracer.startActiveSpan()

naseemkullah:
adds helper function tracer.startActiveSpan()  which creates a new span and activates it in the current Context

### DIFF
--- a/src/trace/NoopTracer.ts
+++ b/src/trace/NoopTracer.ts
@@ -14,8 +14,9 @@
  * limitations under the License.
  */
 
-import { getSpanContext } from '../trace/context-utils';
+import { context } from '../';
 import { Context } from '../context/types';
+import { getSpanContext, setSpan } from '../trace/context-utils';
 import { NonRecordingSpan } from './NonRecordingSpan';
 import { Span } from './span';
 import { isSpanContextValid } from './spancontext-utils';
@@ -44,6 +45,45 @@ export class NoopTracer implements Tracer {
     } else {
       return new NonRecordingSpan();
     }
+  }
+
+  startActiveSpan<F extends (span: Span) => ReturnType<F>>(
+    name: string,
+    arg2: F | SpanOptions,
+    arg3?: F | Context,
+    arg4?: F
+  ): ReturnType<F> | undefined {
+    let fn: F | undefined,
+      options: SpanOptions | undefined,
+      activeContext: Context | undefined;
+    if (arguments.length === 2 && typeof arg2 === 'function') {
+      fn = arg2;
+    } else if (
+      arguments.length === 3 &&
+      typeof arg2 === 'object' &&
+      typeof arg3 === 'function'
+    ) {
+      options = arg2;
+      fn = arg3;
+    } else if (
+      arguments.length === 4 &&
+      typeof arg2 === 'object' &&
+      typeof arg3 === 'object' &&
+      typeof arg4 === 'function'
+    ) {
+      options = arg2;
+      activeContext = arg3;
+      fn = arg4;
+    }
+
+    const parentContext = activeContext ?? context.active();
+    const span = this.startSpan(name, options, parentContext);
+    const contextWithSpanSet = setSpan(parentContext, span);
+
+    if (fn) {
+      return context.with(contextWithSpanSet, fn, undefined, span);
+    }
+    return;
   }
 }
 
--- a/src/trace/ProxyTracer.ts
+++ b/src/trace/ProxyTracer.ts
@@ -38,6 +38,16 @@ export class ProxyTracer implements Tracer {
     return this._getTracer().startSpan(name, options, context);
   }
 
+  startActiveSpan<F extends (span: Span) => unknown>(
+    _name: string,
+    _options: F | SpanOptions,
+    _context?: F | Context,
+    _fn?: F
+  ): ReturnType<F> {
+    const tracer = this._getTracer();
+    return Reflect.apply(tracer.startActiveSpan, tracer, arguments);
+  }
+
   /**
    * Try to get a tracer from the proxy tracer provider.
    * If the proxy tracer provider has no delegate, return a noop tracer.
--- a/src/trace/tracer.ts
+++ b/src/trace/tracer.ts
@@ -37,4 +37,63 @@ export interface Tracer {
    *     span.end();
    */
   startSpan(name: string, options?: SpanOptions, context?: Context): Span;
+
+  /**
+   * Starts a new {@link Span} and calls the given function passing it the
+   * created span as first argument.
+   * Additionally the new span gets set in context and this context is activated
+   * for the duration of the function call.
+   *
+   * @param name The name of the span
+   * @param [options] SpanOptions used for span creation
+   * @param [context] Context to use to extract parent
+   * @param fn function called in the context of the span and receives the newly created span as an argument
+   * @returns return value of fn
+   * @example
+   *   const something = tracer.startActiveSpan('op', span => {
+   *     try {
+   *       do some work
+   *       span.setStatus({code: SpanStatusCode.OK});
+   *       return something;
+   *     } catch (err) {
+   *       span.setStatus({
+   *         code: SpanStatusCode.ERROR,
+   *         message: err.message,
+   *       });
+   *       throw err;
+   *     } finally {
+   *       span.end();
+   *     }
+   *   });
+   * @example
+   *   const span = tracer.startActiveSpan('op', span => {
+   *     try {
+   *       do some work
+   *       return span;
+   *     } catch (err) {
+   *       span.setStatus({
+   *         code: SpanStatusCode.ERROR,
+   *         message: err.message,
+   *       });
+   *       throw err;
+   *     }
+   *   });
+   *   do some more work
+   *   span.end();
+   */
+  startActiveSpan<F extends (span: Span) => unknown>(
+    name: string,
+    fn: F
+  ): ReturnType<F>;
+  startActiveSpan<F extends (span: Span) => unknown>(
+    name: string,
+    options: SpanOptions,
+    fn: F
+  ): ReturnType<F>;
+  startActiveSpan<F extends (span: Span) => unknown>(
+    name: string,
+    options: SpanOptions,
+    context: Context,
+    fn: F
+  ): ReturnType<F>;
 }
--- a/test/proxy-implementations/proxy-tracer.test.ts
+++ b/test/proxy-implementations/proxy-tracer.test.ts
@@ -17,18 +17,18 @@
 import * as assert from 'assert';
 import * as sinon from 'sinon';
 import {
-  ProxyTracerProvider,
-  SpanKind,
-  TracerProvider,
-  ProxyTracer,
-  Tracer,
-  Span,
+  context,
   NoopTracer,
+  ProxyTracer,
+  ProxyTracerProvider,
   ROOT_CONTEXT,
+  Span,
+  SpanKind,
   SpanOptions,
+  Tracer,
+  TracerProvider,
 } from '../../src';
 import { NonRecordingSpan } from '../../src/trace/NonRecordingSpan';
-
 describe('ProxyTracer', () => {
   let provider: ProxyTracerProvider;
   const sandbox = sinon.createSandbox();
@@ -96,6 +96,10 @@ describe('ProxyTracer', () => {
         startSpan() {
           return delegateSpan;
         },
+
+        startActiveSpan() {
+          // stubbed
+        },
       };
 
       tracer = provider.getTracer('test');
@@ -114,6 +118,34 @@ describe('ProxyTracer', () => {
       assert.strictEqual(span, delegateSpan);
     });
 
+    it('should create active spans using the delegate tracer', () => {
+      // sinon types are broken with overloads, hence the any
+      // https://github.com/DefinitelyTyped/DefinitelyTyped/issues/36436
+      const startActiveSpanStub = sinon.stub<Tracer, any>(
+        delegateTracer,
+        'startActiveSpan'
+      );
+
+      const name = 'span-name';
+      const fn = (span: Span) => {
+        try {
+          return 1;
+        } finally {
+          span.end();
+        }
+      };
+      const opts = { attributes: { foo: 'bar' } };
+      const ctx = context.active();
+
+      startActiveSpanStub.withArgs(name, fn).returns(1);
+      startActiveSpanStub.withArgs(name, opts, fn).returns(2);
+      startActiveSpanStub.withArgs(name, opts, ctx, fn).returns(3);
+
+      assert.strictEqual(tracer.startActiveSpan(name, fn), 1);
+      assert.strictEqual(tracer.startActiveSpan(name, opts, fn), 2);
+      assert.strictEqual(tracer.startActiveSpan(name, opts, ctx, fn), 3);
+    });
+
     it('should pass original arguments to DelegateTracer#startSpan', () => {
       const startSpanStub = sandbox.stub(delegateTracer, 'startSpan');
 
@@ -130,6 +162,7 @@ describe('ProxyTracer', () => {
       assert.deepStrictEqual(Object.getOwnPropertyNames(NoopTracer.prototype), [
         'constructor',
         'startSpan',
+        'startActiveSpan',
       ]);
       sandbox.assert.calledOnceWithExactly(startSpanStub, name, options, ctx);
     });
--- a/test/noop-implementations/noop-tracer.test.ts
+++ b/test/noop-implementations/noop-tracer.test.ts
@@ -16,12 +16,13 @@
 
 import * as assert from 'assert';
 import {
+  context,
   NoopTracer,
+  Span,
   SpanContext,
   SpanKind,
-  TraceFlags,
-  context,
   trace,
+  TraceFlags,
 } from '../../src';
 import { NonRecordingSpan } from '../../src/trace/NonRecordingSpan';
 
@@ -55,5 +56,29 @@ describe('NoopTracer', () => {
     assert(span.spanContext().traceId === parent.traceId);
     assert(span.spanContext().spanId === parent.spanId);
     assert(span.spanContext().traceFlags === parent.traceFlags);
+  });
+
+  it('should accept 2 to 4 args and start an active span', () => {
+    const tracer = new NoopTracer();
+    const name = 'span-name';
+    const fn = (span: Span) => {
+      try {
+        return 1;
+      } finally {
+        span.end();
+      }
+    };
+    const opts = { attributes: { foo: 'bar' } };
+    const ctx = context.active();
+
+    const a = tracer.startActiveSpan(name, fn);
+    assert.strictEqual(a, 1);
+
+    const b = tracer.startActiveSpan(name, opts, fn);
+
+    assert.strictEqual(b, 1);
+
+    const c = tracer.startActiveSpan(name, opts, ctx, fn);
+    assert.strictEqual(c, 1);
   });
 });


src/trace/tracer.ts
```diff
@@ -37,4 +37,25 @@ export interface Tracer {
...
+   *     span.setAttribute('key', 'value');
+   *     span.end();
+   *   });
+   */
+  startActiveSpa<T>(
+    name: string,
+    options?: SpanOptions,
+    context?: Context,
+    callback?: (err: Error, span: Span) => T
+  ): T;
```
naseemkullah, line 60: @dyladan does this look right as per https://github.com/open-telemetry/opentelemetry-js-api/pull/14#issuecomment-829501740?

vmarchaud: There is a typo, should be `startActiveSpan`

vmarchaud:
Since this isn't in the spec and we have the problem for the suppress instrumentation system, can we even merge this in the API ?

src/trace/tracer.ts
naseemkullah: Fixed

naseemkullah:
> Since this isn't in the spec and we have the problem for the suppress instrumentation system, can we even merge this in the API

According to spec this functionality may be offered by a separate operation, would this not be it? https://github.com/open-telemetry/opentelemetry-js-api/pull/14#issuecomment-829501740

I'm ignorant to the suppress instrumentation problem though.

vmarchaud:
> I'm ignorant to the suppress instrumentation problem though.

Its our system to avoid exporting loop to "stop" instrumentation from generating spans which is an additional "operation/API" too but we've been asked to either add it to the spec or package it as a seperate npm package.

> According to spec this functionality may be offered by a separate operation, would this not be it? #14 (comment)

Looks like the spec acknowledge it indeed so this should be fine to add

src/trace/tracer.ts
```diff
@@ -37,4 +37,34 @@ export interface Tracer {
...
+   * @param [context] Context to use to extract parent
+   * @param callback called in the context of the span and receives the newly created span as an argument
+   * @returns what is returned by the callback
+   * @example
+   *   const span = tracer.startActiveSpan('op', {}, undefined, (err, span) => {
+   *     span.setAttribute('key', 'value');
+   *     span.end();
+   *   });
+   */
+  startActiveSpan<F extends (err: Error, span: Span) => ReturnType<F>>(
```
vmarchaud, line 55: i'm not sure to see why the callback have arguements here, it should be the same as the `context.with` method (https://github.com/open-telemetry/opentelemetry-js-api/blob/main/src/context/NoopContextManager.ts#L25)

naseemkullah: Thanks I was a little confused about this, PTAL again

src/trace/tracer.ts
```diff
@@ -37,4 +37,35 @@ export interface Tracer {
...
+   * @param [context] Context to use to extract parent
+   * @param fn function called in the context of the span and receives the newly created span as an argument
+   * @returns return value of fn
+   * @example
+   *   const span = tracer.startActiveSpan('op', span => {
+   *     span.setAttribute('key', 'value');
+   *     span.end();
+   *   });
+   */
+  // todo: figure out overloading
```
naseemkullah, line 55: cannot quite figure out overloading here, any tips?

Flarna:
> we have the problem for the suppress instrumentation system

What is the problem regarding suppress instrumentation here? Isn't this API is just a helper around exiting APIs?

src/trace/tracer.ts
```diff
@@ -37,4 +37,34 @@ export interface Tracer {
...
+  ): ReturnType<F>;
+  startActiveSpan?<F extends (span: Span) => ReturnType<F>>(
+    name: string,
+    options: SpanOptions,
+    fn: F
+  ): ReturnType<F>;
+  startActiveSpan?<F extends (span: Span) => ReturnType<F>>(
+    name: string,
+    options: SpanOptions,
+    context: Context,
```
naseemkullah, line 67: I wonder if we need to be able to pass in the context for this helper function, should it not always be the current context?

dyladan:
> > we have the problem for the suppress instrumentation system
> 
> What is the problem regarding suppress instrumentation here? Isn't this API is just a helper around exiting APIs?

No there is no technical problem. I think @vmarchaud was using it as an example of a feature that we implemented without spec and got pushback from the TC.

src/trace/tracer.ts
```diff
@@ -37,4 +37,34 @@ export interface Tracer {
...
+  /**
+   * Starts a new {@link Span} and activates it in the current Context.
+   *
+   * @param name The name of the span
+   * @param [options] SpanOptions used for span creation
+   * @param [context] Context to use to extract parent
+   * @param fn function called in the context of the span and receives the newly created span as an argument
+   * @returns return value of fn
+   * @example
+   *   const span = tracer.startActiveSpan('op', span => {
```
Flarna, line 50: This sample is quite missleading. Doc tells that `startActiveSpan` returns the return value of `fn` - which is `void` here - but it gets assigned to a variable named `span`.

Besides that I think `startActiveSpan` should always return the created span and as a result `fn` should be of type `(span: Span) => void`. Otherwise it's hard for a caller to end the created span.

naseemkullah: Thanks, that was a mistake indeed. re: what `startActiveSpan` should return, what do you think about the span would bing ended  within the fn?

naseemkullah: Thanks, that was a mistake indeed, I've adjusted the example as if the return type continues to be that of the function for the moment, speaking of which : what do you think about the span ending within the fn?

dyladan:
Why is this WIP?

src/trace/tracer.ts
dyladan: I don't think we can assume it will always be active context. Some users even could be managing context manually so there is no active context.

naseemkullah:
> Why is this WIP?

It is not ready to be merged in its current state, and am still requesting comments/direction, e.g.: https://github.com/open-telemetry/opentelemetry-js-api/pull/54#discussion_r625023928

src/trace/tracer.ts
```diff
@@ -37,4 +37,45 @@ export interface Tracer {
...
+   *         code: SpanStatusCode.ERROR,
+   *         message: err.message,
+   *       });
+   *       throw err;
+   *     } finally {
+   *       span.end();
+   *     }
+   *   });
+   */
+  startActiveSpan?<F extends (span: Span) => ReturnType<F>>(
```
dyladan, line 66: All these overloads make the method quite a bit harder to implement. 2nd argument can be a fn or an options, third argument can be an options or a context or undefined. None of our other APIs are structured this way.

src/trace/tracer.ts
Flarna: Async operations are quite common in NodeJs and not all APIs return promises to await them.
Therefore I think this is quite limiting to the API to require that a single function handles a span from start to end.

But sure if we just forward the return values it's up to the user to either end the span within the funciton or return it and let the caller do it later.

src/trace/tracer.ts
naseemkullah: It does indeed though the point of this is to be a helper function to make life easiest as possible for the user (maybe at the cost of overloads), that said:

`tracer.startActiveSpan('name', {}, undefined/context.active(), span => {....})` is not so bad

src/trace/tracer.ts
naseemkullah: Alright thanks for the input, what I don't quite understand is :

``` 
function myFn() {
  const span = tracer.startActiveSpan('myFn', span => {
    return 1;
  });
  span.setStatus({code: SpanStatusCode.OK});
  span.end();
}
```

Will myFn return 1?

src/trace/tracer.ts
naseemkullah: `tracer.startActiveSpan('name', {}, span => {....}, myContext)` is not bad either. but perhaps we could just go with the current args order if we expect context not to always be active context

dyladan: The more I think about it the more I like the overloaded option even if it is more complicated to implement. fn and name are the only two absolutely required args, but having options after the fn could make the code more difficult to read (have to scroll to the end of the block to find out how the span is started). I like having all args before the function, and empty options object `{}` as well as context being just `active()` 99% of the time is also cumbersome for a convenience function.

naseemkullah: Great! That said I have not 100% succeeded in the implementation. (notice `fn!` in the implentation in noop and proxy tracers)

src/trace/ProxyTracer.ts
```diff
@@ -38,6 +39,32 @@ export class ProxyTracer implements Tracer {
...
+    } else if (typeof contextOrFn === 'function') {
+      const span = this.startSpan(
+        name,
+        optionsOrFn,
+        ContextAPI.getInstance().active()
+      );
+      return contextOrFn(span);
+    } else {
+      const span = this.startSpan(name, optionsOrFn, contextOrFn);
+      return fn!(span);
```
naseemkullah, line 64: todo: how to get rid of `!` cc @dyladan

src/trace/tracer.ts
naseemkullah: Alright I am keeping the current approach with an added example of returning the span in the jsdoc since the user can return what they want (span or other).

naseemkullah:
Ready for review.

todo: tests

src/trace/NoopTracer.ts
```diff
@@ -45,6 +46,47 @@ export class NoopTracer implements Tracer {
...
+  ): ReturnType<F>;
+  startActiveSpan<F extends (span: Span) => ReturnType<F>>(
+    name: string,
+    fn: F,
+    context?: Context,
+    options?: SpanOptions
+  ): ReturnType<F> {
+    if (context) {
+      const span = this.startSpan(name, options, context);
+      return fn(span);
```
Flarna, line 73: I guess this will crash if I call it with 3 arguments, e.g. `tracer.startActiveSpan("foo", opts, fn)`

Maybe start checking from last possible argument. Also consider to use e.g. `typeof(options) === 'function'` instead simple truish checks.

Flarna:
I find the name `startActiveSpan` misleading as the span is never set on active context.
```js
tracer.startActiveSpan("name", (span) => {
  assert.strictEqual(getSpan(context.active()), span);
});
```

dyladan:
> I find the name `startActiveSpan` misleading as the span is never set on active context.
> 
> ```js
> tracer.startActiveSpan("name", (span) => {
>   assert.strictEqual(getSpan(context.active()), span);
> });
> ```

IMO it should be made active. This is the whole point of the PR i thought

naseemkullah:
> > I find the name `startActiveSpan` misleading as the span is never set on active context.
> > ```js
> > tracer.startActiveSpan("name", (span) => {
> >   assert.strictEqual(getSpan(context.active()), span);
> > });
> > ```
> 
> IMO it should be made active. This is the whole point of the PR i thought

yes, oversight of mine... bit of an uphill battle with this one

src/trace/NoopTracer.ts
```diff
@@ -45,6 +46,47 @@ export class NoopTracer implements Tracer {
...
+    context: Context,
+    fn: F
+  ): ReturnType<F>;
+  startActiveSpan<F extends (span: Span) => ReturnType<F>>(
+    name: string,
+    fn: F,
+    context?: Context,
+    options?: SpanOptions
+  ): ReturnType<F> {
+    if (context) {
```
obecny, line 71: I would focus only on number of arguments and would limit the declaration to 1 will be easier including the later implementation for example
(writing here without any validation)

```

startActiveSpan<F extends (span: Span) => ReturnType<F>>(
    name: string,
    arg2: F | options ,
    arg3?: F | Context,
    arg4?: F
  ): ReturnType<F>;

name = arguments[0]
if (arguments.length === 2) {
  callback = arg2;
} else if (arguments.length === 3) {
  options = arg2;
  callback = arg3;
} else if (arguments.length === 4) {
  options = arg2;
  context = arg3;
  callback = arg4;
}

```

I would then skip the validation. TS will be validated automatically, but if user enforce it will be simply a valid failure.

src/trace/NoopTracer.ts
```diff
@@ -45,6 +46,47 @@ export class NoopTracer implements Tracer {
...
+    options: SpanOptions,
+    fn: F
+  ): ReturnType<F>;
+  startActiveSpan<F extends (span: Span) => ReturnType<F>>(
+    name: string,
+    options: SpanOptions,
+    context: Context,
+    fn: F
+  ): ReturnType<F>;
+  startActiveSpan<F extends (span: Span) => ReturnType<F>>(
```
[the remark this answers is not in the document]
naseemkullah: Agreed, tbh i only added this variant for the compiler, i need to figure out the overloads still. But yes, `fn` should always be last

src/trace/NoopTracer.ts
naseemkullah: great thanks will try this shortly

naseemkullah: so TS was not happy without further validation or type assertions, PTAL.

naseemkullah:
Two questions remain before I'm on to tests:

- Should the context that is passed to `startSpanAsActive()`, be both the context in which to set the span as active (1st arg to `setSpan()`), as well as the context used to extract a parent (3rd arg to `startSpan()`)?

- Implementation in noop and proxy tracers are the exact same,  could this be factored out?

src/trace/NoopTracer.ts
```diff
@@ -45,6 +46,46 @@ export class NoopTracer implements Tracer {
...
+      typeof arg4 === 'function'
+    ) {
+      options = arg2;
+      context = arg3;
+      fn = arg4;
+    }
+
+    const ctx = context ?? ContextAPI.getInstance().active();
+    const span = this.startSpan(name, options, ctx);
+    setSpan(ctx, span);
```
naseemkullah, line 81: Should the context that is passed to startSpanAsActive(), be both the context in which to set the span as active (1st arg to setSpan()), as well as the context used to extract a parent (3rd arg to startSpan())?

src/trace/NoopTracer.ts
obecny: you will need to cast them right ?.
```
context = arg3 as Context
callback = arg4 as F;
```
etc.

naseemkullah: Right so the options to make compilée happy were either casting or validation with the "type of" checks, I went with latter since now types are more real

src/trace/ProxyTracer.ts
```diff
@@ -38,6 +40,46 @@ export class ProxyTracer implements Tracer {
     return this._getTracer().startSpan(name, options, context);
   }
 
+  startActiveSpan<F extends (span: Span) => ReturnType<F>>(
+    name: string,
+    arg2: F | SpanOptions,
+    arg3?: F | Context,
+    arg4?: F
+  ): ReturnType<F> {
+    let fn: F | undefined,
```
vmarchaud, line 49: proxy just calls the underlying provider, not re-implement this logic

naseemkullah: thanks! done... I had to coerce args and end with an error throw to compile, PTAL

src/trace/ProxyTracer.ts
```diff
@@ -38,6 +38,24 @@ export class ProxyTracer implements Tracer {
...
 
+  startActiveSpan?<F extends (span: Span) => ReturnType<F>>(
+    name: string,
+    arg2: F | SpanOptions,
+    arg3?: F | Context,
+    arg4?: F
+  ): ReturnType<F> {
+    const tracer = this._getTracer();
+    if (tracer.startActiveSpan) {
+      return tracer.startActiveSpan(
```
Flarna, line 49: This wont work if the implementation in delegate is done like in NoopTracer because the delegate will always see `arguments.length === 4`. If only two are passed the remaining two are set then as `undefined`.
I guess `Reflect.apply(tracer.startActiveSpan, tracer, arguments)` should work.

src/trace/ProxyTracer.ts
```diff
@@ -38,6 +38,24 @@ export class ProxyTracer implements Tracer {
...
+        name as string,
+        arg2 as SpanOptions,
+        arg3 as Context,
+        arg4 as F
+      );
+    }
+    throw new Error('startActiveSpan not implemented in delegate');
```
Flarna, line 56: Usually we don't throw in API as far as I know. For other APIs no checks are done.

src/trace/NoopTracer.ts
Flarna: I think so. But what is missing here is to set the span active. `setSpan()` just returns a new context which is never used here.
it should be something like `api.context.with(setSpan(ctx, span), fn)`

src/trace/ProxyTracer.ts
vmarchaud: We can't indeed

naseemkullah: ok i will try adding `| undefined` to return type

src/trace/NoopTracer.ts
naseemkullah: Ah right, thanks!

src/trace/ProxyTracer.ts
Flarna: Why? for `startSpan` we just forward and return the result. We should do the same here.

Clearly this adds a requirement to all SDKs out somewhere as we add a new public API to Tracer which needs to be implemented by everyone to stay compatible with the API.

naseemkullah: without `| undefined` I get

image

worth noting `startActiveSpan` is being added as an optional method atm (thats why there is  a `if (tracer.startActiveSpan)` check)

src/trace/ProxyTracer.ts
naseemkullah: wow this worked, ill have to look up Reflect later!

src/trace/ProxyTracer.ts
dyladan: `startActiveSpan` shouldn't need to be implemented by the SDK. it is completely implementable with api methods only.

(no editor help here so this might not work exactly, but it illustrates the point.

```typescript
class TraceAPI {
  // ...


  public startActiveSpan(name, options, context, cb, ...args) {
    const span = this.startSpan(name, options, context);
    context.with(setSpan(context, span), cb, ...args)
  }
```

src/trace/NoopTracer.ts
naseemkullah: struggling with

image

src/trace/ProxyTracer.ts
Flarna: @dyladan I fear the concrete `Tracer` instance is missing here. `TraceApi#startSpan` doesn't exist.

`function startActiveSpan(tracer, name,...)` as free helper function would work.

src/trace/NoopTracer.ts
Flarna: maybe `api.context.with(setSpan(ctx, span), fn, undefined, span)` as the argument after `fn` is the `thisArg` and then the real args come.

naseemkullah: thank you

naseemkullah: I've made this change and changed a few var names, PTAL if the new var names make sense: `contextSingleton`, `activeContext`/`activeCtx`

src/trace/tracer.ts
```diff
@@ -37,4 +37,60 @@ export interface Tracer {
...
+   *         code: SpanStatusCode.ERROR,
+   *         message: err.message,
+   *       });
+   *       throw err;
+   *     }
+   *   });
+   *   do some more work
+   *   span.end();
+   */
+  startActiveSpan?<F extends (span: Span) => ReturnType<F>>(
```
dyladan, line 81: Why is this in the interface? It can be implemented entirely in API methods

naseemkullah: I looked for instances of `startSpan` and went from there, with this new understanding i will remove it then

naseemkullah: ooc what is the negative consequence of having it as optional in the interface?

vmarchaud: > Why is this in the interface? It can be implemented entirely in API methods

Are you suggesting to move it into the trace api directly ? I believe it can't since it needs a tracer to do it.

naseemkullah: if `tracer.startActiveSpan(name, fn)` is the ux we are trying to provide, it should be defined in the interface if I'm not mistaken, but if we want to now change that to `startActiveSpan(tracer, spanName, fn)` that is possible purely in trace api, though I prefer the former

Flarna: Why? It's only needed to add the tracer to use as argument.
Creating a tracer would require a tracer name + version.

MSNev (approved):

src/trace/ProxyTracer.ts
```diff
@@ -38,6 +38,17 @@ export class ProxyTracer implements Tracer {
     return this._getTracer().startSpan(name, options, context);
   }
 
+  startActiveSpan?<F extends (span: Span) => ReturnType<F>>():
+    | ReturnType<F>
+    | undefined {
+    const tracer = this._getTracer();
+
+    if (tracer.startActiveSpan) {
```
vmarchaud, line 46: since we added this into the tracer interface, `startActiveSpan` should always be defined right ?

naseemkullah: It's currently optional, if we settle on doing this in the interface should it be required?

naseemkullah: I know @vmarchaud proposed that it should be required. If any else wants to chime in though

vmarchaud: Yeah actually i though it was already required, we discussed this at the SIG (was it you that added it into our agenda ?) and settle into require it into the interface.

naseemkullah: Ah great! Yes I added it, thanks for discussing and confirming! 👍

src/trace/tracer.ts
naseemkullah: According to SIG notes, this will be added in interface and no base class for now. Thanks to SIG attendees, resolving!

src/trace/ProxyTracer.ts
naseemkullah: Alright, with `startActiveSpan` being required in the interface, no more need to return `| undefined`.

src/trace/tracer.ts
dyladan: To expand on why we decided no base class, we previously had classes and base classes in the API and they were causing versioning conflicts even when the public interface was not changed. Typescript doesn't consider two classes to be the same unless _all_ of their properties have the same type and changing private properties causes type errors. The effort of hiding classes behind interfaces is possible but difficult to enforce in third party code and an easy mistake to make. Basically the benefit isn't big enough to risk versioning conflicts in the future.

vmarchaud (approved):
lgtm, tests are failing though

test/proxy-implementations/proxy-tracer.test.ts
```diff
@@ -96,6 +98,50 @@ describe('ProxyTracer', () => {
         startSpan() {
           return delegateSpan;
         },
+
+        startActiveSpan<F extends (span: Span) => ReturnType<F>>(
```
Flarna, line 102: Is it really needed for the test to implememt a full tracer here? What about using a `Sinon.Stub` and verify that it is called with the right set of arguments?

Flarna (approved):
A test for `NoopTracer` to verify that fn is actually called correct and return value is passed through would be nice.

obecny (approved):
lgtm

src/trace/NoopTracer.ts
```diff
@@ -45,6 +46,50 @@ export class NoopTracer implements Tracer {
...
+      typeof arg2 === 'object' &&
+      typeof arg3 === 'object' &&
+      typeof arg4 === 'function'
+    ) {
+      options = arg2;
+      activeContext = arg3;
+      fn = arg4;
+    }
+
+    const contextSingleton = ContextAPI.getInstance();
```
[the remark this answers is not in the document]
naseemkullah: 💯 thanks

test/proxy-implementations/proxy-tracer.test.ts
```diff
@@ -114,6 +154,12 @@ describe('ProxyTracer', () => {
       assert.strictEqual(span, delegateSpan);
     });
 
+    it('should create active spans using the delegate tracer', () => {
+      tracer.startActiveSpan('test', span => {
+        assert.strictEqual(getSpan(context.active()), span);
```
naseemkullah, line 159: seems like th eactive context at this point is not the one with the span set on it, investigating... same for the noop-tracer test

Flarna: This is expected, the NoopContextManager always returns ROOT_CONTEXT. see #23

naseemkullah: ah, thanks!

src/trace/ProxyTracer.ts
```diff
@@ -38,6 +38,11 @@ export class ProxyTracer implements Tracer {
     return this._getTracer().startSpan(name, options, context);
   }
 
+  startActiveSpan<F extends (span: Span) => ReturnType<F>>(): ReturnType<F> {
+    const tracer = this._getTracer();
+    return Reflect.apply(tracer.startActiveSpan, tracer, arguments);
```
dyladan, line 43: Why are you using Reflect.apply here instead of just calling the function like above in startSpan?

test/proxy-implementations/proxy-tracer.test.ts
naseemkullah: Trying to implement, seems to have caught a bug, 4 args are always expected 

image

src/trace/ProxyTracer.ts
naseemkullah: it was to get rid of the `| undefined` in the return type if I recall correctly

naseemkullah: please see https://github.com/open-telemetry/opentelemetry-js-api/pull/54#discussion_r626596133

naseemkullah: could do this if preferred?

```js
  if (arg4) {
      return this._getTracer().startActiveSpan(name, arg2, arg3, arg4);
    } else if (arg3) {
      return this._getTracer().startActiveSpan(name, arg2, arg3);
    }
    return this._getTracer().startActiveSpan(name, arg2);
    ```

Comment (dyladan) [reply src/trace/ProxyTracer.ts:43]: nah just wanted to know

File: src/trace/tracer.ts [hunk 21]
```diff
@@ -37,4 +37,63 @@ export interface Tracer {
...
+  startActiveSpan<F extends (span: Span) => ReturnType<F>>(
+    name: string,
+    fn: F
+  ): ReturnType<F>;
+  startActiveSpan<F extends (span: Span) => ReturnType<F>>(
+    name: string,
+    options: SpanOptions,
+    fn: F
+  ): ReturnType<F>;
+  startActiveSpan<F extends (span: Span) => ReturnType<F>>(
```
dyladan, line 93: Your `F` generic doesn't need to specify the return type of F is `ReturnType<F>` (this is a truism). We don't care what it is so you can just make it unknown.

```suggestion
  startActiveSpan<F extends (span: Span) => unknown>(
    name: string,
    fn: F
  ): ReturnType<F>;
  startActiveSpan<F extends (span: Span) => unknown>(
    name: string,
    options: SpanOptions,
    fn: F
  ): ReturnType<F>;
  startActiveSpan<F extends (span: Span) => unknown>(
```

src/trace/ProxyTracer.ts
```diff
@@ -38,6 +38,11 @@ export class ProxyTracer implements Tracer {
     return this._getTracer().startSpan(name, options, context);
   }
 
+  startActiveSpan<F extends (span: Span) => ReturnType<F>>(): ReturnType<F> {
```
dyladan, line 41: Your sinon stub is broken because it can't inspect the arguments of this type I believe.

```suggestion
  startActiveSpan<F extends (span: Span) => unknown>(
    _name: string,
    _options: F | SpanOptions,
    _context?: F | Context,
    _fn?: F
  ): ReturnType<F> {
```

dyladan (commented):
Hopefully the comment I left can help you with your sinon stub typing

naseemkullah:
> Hopefully the comment I left can help you with your sinon stub typing

Thanks! 🙏

naseemkullah:
> Hopefully the comment I left can help you with your sinon stub typing

In the end it was still broken, sinon types + overloads don't play nicely as per https://github.com/DefinitelyTyped/DefinitelyTyped/issues/36436

But was able to work around it with `any`. I am going to now squash my commits, this PR is finally good to go

dyladan (approved):

naseemkullah:
Thanks for the help everyone!

PR #55, opened 2021-05-02
chore: move baggage methods in propagation namespace

vmarchaud:
Since we moved the span context method into `trace` namespace, i'm also proposing to move `[create/get/set]Baggage` into the `propagation` namespace.

We still export `baggageEntryMetadataFromString` which feels kinda wrong to me since it doesn't do anything apart creating a unusued object, could we have a less strict type for the metadata so we don't need this function ?


obecny (commented):
Because the methods will be available under propagation namespace instead of global, i think you should add info to `Upgrade Guidelines`, other than that lgtm

dyladan (approved):
I agree with bart you should update the upgrade guidelines. Other than that LGTM so I will approve

vmarchaud:
@obecny I rebased + added the upgrade guideline, PTAL

vmarchaud:
@obecny I rebased + added the upgrade guideline, PTAL

obecny (approved):
lgtm

PR #58, opened 2021-05-03
chore: use spancontext for link

### LINKED ISSUE #57: Use whole spancontext for link
From the spec:

>A Link is structurally defined by the following properties:
>
>- SpanContext of the Span to link to.
>- Zero or more Attributes further describing the link.

Seems pretty clear to me that the link should expose the whole SpanContext including the trace state.

_Originally posted by @dyladan in https://github.com/open-telemetry/opentelemetry-js-api/issues/36#issuecomment-831239711_

Currently our `Link` is an object that shares some properties with `SpanContext`. Should we (1) add the missing properties to `Link` or should we (2) make it `type Link = { spanContext: SpanContext, attributes: Attributes }`?

To me, the spec seems to imply 2 is the more "correct" choice, but 1 is the less breaking change.

dyladan:
Fixes #57


blumamir (approved):

naseemkullah (approved):

src/trace/link.ts
```diff
@@ -34,7 +34,7 @@ import { LinkContext } from './link_context';
  */
 export interface Link {
   /** The {@link LinkContext} of a linked span. */
```
blumamir, line 36: Update the comment maybe? `LinkContext` -> `SpanContext`

src/trace/link.ts
```diff
@@ -33,8 +33,8 @@ import { LinkContext } from './link_context';
  *    Service Provider) can be correlated.
  */
 export interface Link {
-  /** The {@link LinkContext} of a linked span. */
-  context: LinkContext;
+  /** The {@link SpanContext} of a linked span. */
+  context: SpanContext;
```
Flarna, line 37: by using `SpanContext` not only `traceState` is added optional, it adds also `traceFlags` mandatory. 
Is this intended?

dyladan: The spec states that a link contains "SpanContext of the Span to link to."

I think the most common way to get the spancontext will be to call `span.spanContext()` so the user won't usually be responsible for manually creating an object and ensuring it is a valid spancontext anyway.

I can change it to just include the properties from the Proto if you think it would be better. Maybe the spec should also be clarified here.

Flarna: I think it's ok to keep it. Moving it top optional later would be non breaking.

Flarna (approved):

dyladan:
Created https://github.com/open-telemetry/opentelemetry-specification/issues/1667

vmarchaud (approved):

obecny (approved):
lgtm

Issue #61, opened 2021-04-12
Namespaces in diag / logger


obecny:
I'm not sure if such issue has been already created, but I couldn't find it.
Because we did major cleanup with plugins and other stuff, we can now go back to namespacing in diag which would give us a better understanding with logs.
I would like to implement following pattern in few places - with instrumentation it will be the easiest.
Each instrumentation calls a constructor with name. We will use this name to create a logger class that will be already namespaced. So in all instrumentations we will just have one protected logger which we will use instead of api.diag.
This way the logs will be transparent and it will look like 
`this.diag.error('foo')` 
// will show
// "@opentelemetry/instrumentation-http", "foo"

under the hood those loggers will always be called diag but with one extra param in front of it

and  new public`diag` property will be created automatically in constructor.

We can discuss if it should be public or protected, but I would vote for public to be able to call it easily from any subclass etc.

For other things like exporter we can add things after this change to avoid too many changes and overhead.

The first part will update instrumentations in core. After release I will update contrib and meanwhile I believe we can already have agreement towards exporters and anything else that we think should have it.

I don't want to start before I have some thumbs up for it.

PR #62, opened 2021-05-06
chore: adding component logger

obecny:
Adding component logger that will support namespace. New function to diag `createComponentLogger`.

Fixes https://github.com/open-telemetry/opentelemetry-js-api/issues/61


src/diag/ComponentLogger.ts
```diff
@@ -0,0 +1,70 @@
...
+  nameSpace: string,
+  args: any
+): void {
+  const logger = getGlobal('diag');
+  // shortcut if logger not set
+  if (!logger) {
+    return;
+  }
+
+  args.unshift(nameSpace);
```
naseemkullah, line 68:
```suggestion
  namespace: string,
  args: any
): void {
  const logger = getGlobal('diag');
  // shortcut if logger not set
  if (!logger) {
    return;
  }

  args.unshift(namespace);
```
nit

naseemkullah (approved):

src/diag/types.ts
```diff
@@ -89,3 +89,49 @@ export enum DiagLogLevel {
...
+   * in a production environment.
+   */
+  verbose: (...args: any) => void;
+}
+
+/**
+ * Defines options for ComponentLogger
+ */
+export interface ComponentLoggerOptions {
+  namespace?: string;
```
vmarchaud, line 136: should `namespace` be mandatory ?

src/diag/ComponentLogger.ts
obecny: I was starring for 1 minute trying to figure out the change :D, nice catch, thx

src/diag/types.ts
vmarchaud: i think i would prefer it to be mandatory, since we are adding it so component logs can be more precise i wouldn't make sense for them to not use it

src/diag/types.ts
```diff
@@ -89,3 +89,49 @@ export enum DiagLogLevel {
   /** Used to set the logging level to include all logging */
   ALL = 9999,
 }
+
+/**
+ * Defines options for ComponentLogger
+ */
+export interface ComponentLogger {
```
dyladan, line 96: Does this need its own interface? There shouldn't be a need to read the namespace after getting the logger and the other methods just come from DiagLogger right?

src/diag/ComponentLogger.ts
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Component Logger which is meant to be used as part of any component which
+ * will add automatically additional namespace in front of the log message.
+ * It will then forward all message to global diag logger
+ * @example
+ * const cLogger = diag.createComponentLogger({ namespace: '@opentelemetry/instrumentation-http' });
+ * cLogger.debug('test');
+ * // @opentelemetry/instrumentation-http test
+ */
+export class DiagComponentLogger implements ComponentLogger {
```
dyladan, line 29: I think this should implement DiagLogger and make namespace private

src/diag/types.ts
obecny: yes, it needs the new interface, the functions are a bit different, if you look closer

vmarchaud (approved):

dyladan (approved):

PR #63, opened 2021-05-11
feat: debug log global registrations and logger overwrites

Rauno56:
Currently, there's no visibility to global registration which is a relatively important operation.
For example, there's no way to know if a logger has been overwritten or not which might mean the successive log messages/errors from other parts of the API might not be appearing without the developer's knowledge.

To relieve that, a few log statements were added:

- There can be only one global logger. If an existing one is overwritten, *both* of the loggers get a `warn` log line about the event.
- Any time a global is (un)registered, `debug` log line is created with the type and version.

It relieves the immediate pain described in https://github.com/open-telemetry/opentelemetry-js-api/issues/56


dyladan (approved):

dyladan:
Really like this idea. Can be frustrating for users when they think they are loading a diag logger but it gets overwritten. Maybe it would be even worth getting the stack when it is overwritten?

```typescript
// use empty string as fallback in case stack can't be generated
const stack = new Error().stack ?? ""

oldLogger.warn(`A new DiagLogger was registered which overwrites this logger from ${stack}`);
```

Rauno56:
Having a stack is a really good idea as well. I'll add that here for now to keep the focus of the PR.

Although... whenever I see variables in log messages I see a lost opportunity to structure the log entry:

1. Thought's on using structured logging everywhere instead? An API similar to `pino` perhaps:

```
log(object, string);
log(object | string);
```

2. What do you think of making it an option in the logger itself? `logStackTrace: <LEVEL>/boolean`.

test/internal/global.test.ts
```diff
@@ -91,48 +100,47 @@ describe('Global Utils', () => {
...
+    api1.diag.setLogger(logger1);
+    api1.diag.setLogger(logger2);
+
+    sinon.assert.notCalled(logger1.error);
+    sinon.assert.calledOnce(logger1.warn);
+    sinon.assert.calledWith(logger1.warn, sinon.match(/will be overwritten/i));
+
+    sinon.assert.notCalled(logger2.error);
+    sinon.assert.calledOnce(logger2.warn);
+    sinon.assert.calledWith(logger2.warn, sinon.match(/will overwrite/i));
```
johnbley, line 144: ... and then `logger2.info('check that logger2 is actually now the active log');` with the appropriate assert?

johnbley (approved):

test/internal/global.test.ts
Rauno56: I assumed one of the ~500 tests that test logger does that already. I was wrong and TBH et probably belongs to global tests anyways. Added it. Thanks!

test/diag/logLevel.test.ts
```diff
@@ -161,35 +164,37 @@ describe('LogLevelFilter DiagLogger', () => {
...
-                  ) {
-                    assert.deepStrictEqual(calledArgs[lName], [
-                      `${fName} called %s`,
-                      'param1',
-                    ]);
-                  } else {
-                    assert.strictEqual(calledArgs[lName], null);
-                  }
-                });
+          it('diag.setLogger level is ignored when appropriate', () => {
```
obecny, line 167: Previous version was much more readable you could see all tests that were failing or not, after this it will fail on first and nothing else will be shown, I would rather revert that

src/internal/global-utils.ts
```diff
@@ -35,13 +35,12 @@ export function registerGlobal<Type extends keyof OTelGlobalAPI>(
   instance: OTelGlobalAPI[Type],
   allowOverride = false
 ): boolean {
-  _global[GLOBAL_OPENTELEMETRY_API_KEY] = _global[
+  const api = (_global[GLOBAL_OPENTELEMETRY_API_KEY] = _global[
```
obecny, line 38: this is hard for readability, it takes a bit time to realise what is what, I wouldn't squeeze too many things at once.

vmarchaud (approved):

src/internal/global-utils.ts
Rauno56: I didn't do this for saving keystrokes but to make it obvious that `api` has a default value and cannot be `undefined`.

test/diag/logLevel.test.ts
Rauno56: The previous version had `describe` and `it` rows for every assertion.
The reason I made it a little bit more concise was exactly that the flood of errors didn't help me to pinpoint the issue when I broke the tests implementing it.

Even though it's hard for _me_ to imagine a case where the previous version would actually come in handy doesn't mean there isn't one - could you please give me an example perhaps?

dyladan: Sometimes I am overwhelmed by failures too and tend to use `-b` flag while debugging issues for this reason. 

That said, personally I prefer the old style with many, more specific, tests. One large test which tests everything can be difficult to debug for other reasons. If I get 4 errors and see that all of them happen when i'm on `INFO` level, then it gives me something to look for. If everything happens in a single test with the name 'when appropriate', then i have no possibility of making those correlations.

In short, I agree with @obecny

obecny: that is also the main reason it is called unit test. It tests one particular case. If it fails it tells you exactly what has happened by just reading the description, you don't need to even check the code or test as it will be self explained. And with all the tdd approach as well you have basically structure
```
when A is 0
  then B is 0
  AND C is 0
    then D is 0
    then E is 2
```

this way you have self explained logic of your code imho the best possible way of any1 reading your code and who want to change it or do refactoring. And by writing the tests like this you will immediately know if your code has cyclomatic complexity that is too complex or it is fine.
Of course it is not like this in all of our code, might not always be possible etc., but that would be desired way at least this is what I think :)

Rauno56: > If everything happens in a single test with the name 'when appropriate', ...

If that was the case, for sure! I moved the description from the test name into the assertion - whenever the test fails, it doesn't just say "when appropriate", but a message like the one that was in the test name + declaring the expectation.
See lines [182](https://github.com/open-telemetry/opentelemetry-js-api/pull/63/files/479286e911494a0da55aa5da7288cd36065e7a02#diff-d54b8a1af38e0facfcb2c0de13d09e036f342c7416ea7ab9cd0870856695d5b7R182) and [188](https://github.com/open-telemetry/opentelemetry-js-api/pull/63/files/479286e911494a0da55aa5da7288cd36065e7a02#diff-d54b8a1af38e0facfcb2c0de13d09e036f342c7416ea7ab9cd0870856695d5b7R188).

> that is also the main reason it is called unit test [...]

I feel the argumentation is starting to get more about
- semantics(unforgivingly :) paraphrasing how I understood your point: "I call it [tdd/a unit test], therefore, it must follow the strict definition"),
- style(assumptions about "anyone" and everyone even though I'm clearly not one of them) and
- unrelated topics("cyclomatic complexity"),

and less about the noise-to-signal ratio I intended to improve which perhaps proves that it's not as effective as I hoped!

src/internal/global-utils.ts
dyladan: I actually like that this removes the `!` assertion. In the old version, typescript thought `api` could still be undefined when in fact it couldn't. @obecny are you OK resolving this or do you feel strong here?

test/diag/logLevel.test.ts
Rauno56: I'll revert those changes.

src/internal/global-utils.ts
obecny: sounds good

Rauno56:
@dyladan, @obecny. Reverted the changes to test structure. Everything should now be addressed.

obecny (approved):
lgtm, thx for changes

PR #68, opened 2021-05-14
Add ability to see attributes

PepijnSenders:
This PR relies on: https://github.com/open-telemetry/opentelemetry-js/pull/2207

### DIFF
--- a/src/trace/span.ts
+++ b/src/trace/span.ts
@@ -62,6 +62,18 @@ export interface Span {
   setAttributes(attributes: SpanAttributes): this;
 
   /**
+   * Get the attribute of the Span by key.
+   * 
+   * @param key Key of the span attribute
+   */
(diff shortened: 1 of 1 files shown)


vmarchaud:
I'm not sure we can this into the API since this is not in the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md)

dyladan:
> I'm not sure we can this into the API since this is not in the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md?rgh-link-date=2021-05-14T17%3A03%3A47Z)

Not only is it not in the spec, it is specifically denied by the spec.

From spec:

> The API-level definition for Span's interface only defines write-only access to the span. This is good because instrumentations and applications are not meant to use the data stored in a span for application logic.

PR #69, opened 2021-05-15
feat(context): add utils method to remove keys from context #66

### LINKED ISSUE #66: improv: add context utils to delete span/baggage from context
Maybe we should add `deleteSpan(context)` (and `deleteBaggage(context)`) at some time?

_Originally posted by @Flarna in https://github.com/open-telemetry/opentelemetry-js/pull/2185#discussion_r631128874_

vmarchaud:
Ref: 
- https://github.com/open-telemetry/opentelemetry-js-api/issues/66
- https://github.com/open-telemetry/opentelemetry-js/pull/2185#discussion_r631128874

Fixes #66

### DIFF
--- a/src/api/trace.ts
+++ b/src/api/trace.ts
@@ -28,7 +28,6 @@ import { Tracer } from '../trace/tracer';
 import { TracerProvider } from '../trace/tracer_provider';
 import {
   deleteSpan,
-  deleteSpanContext,
   getSpan,
   getSpanContext,
   setSpan,
@@ -92,8 +91,6 @@ export class TraceAPI {
   public isSpanContextValid = isSpanContextValid;
 
   public deleteSpan = deleteSpan;
-
-  public deleteSpanContext = deleteSpanContext;
 
   public getSpan = getSpan;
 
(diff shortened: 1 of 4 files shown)


src/api/trace.ts
```diff
@@ -89,6 +91,10 @@ export class TraceAPI {
 
   public isSpanContextValid = isSpanContextValid;
 
+  public deleteSpan = deleteSpan;
+
+  public deleteSpanContext = deleteSpanContext;
```
Flarna, line 96: I see no real benefit in `delteSpanContext()`.
`setSpanContext()` is only a short form to for `setSpan()` therefore `deleteSpan()` should be fine IMHO.

vmarchaud: Well having those shortcuts (which by definition can be done otherwise) basically have no cost to have at the top level like this, so why not have them ?

Flarna: mainly because delete doesn't delete anything here - it creates a new `Span`.

Deleting a `SpanContext` is in my opinion the same as deleting a `Span` as a `Span` without `SpanContext` is no longer a `Span`

dyladan: I agree with @Flarna. Having `deleteSpanContext` set a new span with invalid span context isn't obvious to me at all.

vmarchaud: I will change this then, ty for feedback

Flarna (approved):

dyladan (approved):

obecny (approved):
lgtm

PR #71, opened 2021-05-18
chore: export baggage

dyladan:
Baggage types were not exported


src/baggage/index.ts
```diff
@@ -14,66 +14,6 @@
...
-  }
-
-  return {
-    __TYPE__: baggageEntryMetadataSymbol,
-    toString() {
-      return str;
-    },
-  };
-}
+export * from './context-helpers';
```
vmarchaud, line 17: we shouldn't export context helpers or `createBaggage` a the top level, so we should remove this line and line 18 should be `export { baggageEntryMetadataFromString } from './types';`

dyladan: Why would we not export them? Context helpers just get and set baggage on the context, an operation people may reasonably want to perform. How do you access baggage without them?

vmarchaud: Because they are exposed on the `propagation` namespace, that was the point of #55

dyladan: Latest commit look ok then?

obecny, line 17: hmm this line shouldn't be exported ?

obecny (approved):
I think it looks fine now, just wait for confirmation yet from @vmarchaud yet

vmarchaud (approved):

src/baggage/index.ts
vmarchaud: :+1:

dyladan: Maybe we should remove the index.ts file entirely and reference things directly to remove confusion? Or we could move the context helpers into the `internal` folder to show they should not be exported.

PR #90, opened 2021-06-05
chore: enable typescript 4.3 noImplicitOverride option

Flarna:
Enable typescript 4.3 feature noImplicitOverride to improve code quality.
Besides that some other dev-dependencies are updated.


vmarchaud (commented):
shouldn't all noop implementation use `override` keyword too ?

Flarna:
> shouldn't all noop implementation use `override` keyword too ?

The noop classes do not override anything from a base class - they implement an interface. `override` in typescript seems to be slightly different then in java or C++.

vmarchaud (approved):

dyladan (approved):

PR #94, opened 2021-06-09
chore: enforce strict equality on prerelease versions

dyladan:
This PR prevents the API from accidentally accepting compatibility from e.g. `1.0.0` and `1.0.0-rc.3`.

It is a requirement if we want the possibility of ever using canary releases.


src/internal/semver.ts
```diff
@@ -50,8 +50,16 @@ export function _makeCompatibilityCheck(
     major: +myVersionMatch[1],
     minor: +myVersionMatch[2],
     patch: +myVersionMatch[3],
+    prerelease: myVersionMatch[4],
   };
 
+  // if ownVersion has a prerelease tag, versions must match exactly
+  if (ownVersionParsed.prerelease) {
```
Flarna, line 57:
```suggestion
  if (ownVersionParsed.prerelease != null) {
```

src/internal/semver.ts
```diff
@@ -16,7 +16,7 @@
 
 import { VERSION } from '../version';
 
-const re = /^(\d+)\.(\d+)\.(\d+)(?:-(.*))?$/;
+const re = /^(\d+)\.(\d+)\.(\d+)(-(.*))?$/;
```
Flarna, line 19: maybe this should be `/^(\d+)\.(\d+)\.(\d+)(-(.+))?$/` to disallow empty dev versions?

Flarna (approved):

vmarchaud (approved):

obecny (approved):

PR #97, opened 2021-06-23
feat: allow setting attributes when recording exceptions

### LINKED ISSUE #95: Span.recordException must allow additional attributes according to spec
Not sure if this ought to be labelled a "bug" or "discussion". I noticed a discrepancy between [span.recordException](https://github.com/open-telemetry/opentelemetry-js/blob/7750282034ee58bdd257446c2970f01307d7522a/packages/opentelemetry-tracing/src/Span.ts#L189) and the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception)

Specifically, this part (current implementation does not allow passing additional attributes):
> If RecordException is provided, the method MUST accept an optional parameter to provide any additional event attributes (this SHOULD be done in the same way as for the AddEvent method). If attributes with the same name would be generated by the method already, the additional attributes take precedence.

- [x] This only affects the JavaScript OpenTelemetry library
- [ ] This may affect other libraries, but I would like to get opinions here first

vreynolds:
* span.recordException needs to allow settings event attributes per [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#record-exception)
> If RecordException is provided, the method MUST accept an optional parameter to provide any additional event attributes (this SHOULD be done in the same way as for the AddEvent method). If attributes with the same name would be generated by the method already, the additional attributes take precedence.
* Fixes #95

### DIFF
--- a/src/trace/span.ts
+++ b/src/trace/span.ts
@@ -65,14 +65,21 @@ export interface Span {
    * Adds an event to the Span.
    *
    * @param name the name of the event.
-   * @param [attributesOrStartTime] the attributes that will be added; these are
-   *     associated with this event. Can be also a start time
-   *     if type is {@type TimeInput} and 3rd param is undefined
+   * @param [startTime] start time of the event.
+   */
+  addEvent(name: string, startTime?: TimeInput): this;
+
+  /**
+   * Adds an event to the Span.
+   *
+   * @param name the name of the event.
+   * @param [attributes] the attributes that will be added; these are
+   *     associated with this event.
    * @param [startTime] start time of the event.
    */
   addEvent(
     name: string,
-    attributesOrStartTime?: SpanAttributes | TimeInput,
+    attributes?: SpanAttributes,
     startTime?: TimeInput
   ): this;
 
@@ -123,15 +130,22 @@ export interface Span {
    * Sets exception as a span event.
    *
    * @param exception the exception the only accepted values are string or Error
-   * @param [attributesOrTime] additional attributes to be associated with
-   *     this event. Can also be the time parameter if type is {@type TimeInput}
-   *     and 3rd parameter is undefined.
+   * @param [time] the time to set as Span's event time. If not provided,
+   *     use the current time.
+   */
+  recordException(exception: Exception, time?: TimeInput): void;
+
+  /**
+   * Sets exception as a span event.
+   *
+   * @param exception the exception the only accepted values are string or Error
+   * @param [attributes] additional attributes to be associated with this event.
    * @param [time] the time to set as Span's event time. If not provided,
    *     use the current time.
    */
   recordException(
     exception: Exception,
-    attributesOrTime?: SpanAttributes | TimeInput,
+    attributes?: SpanAttributes,
     time?: TimeInput
   ): void;
 }
--- a/src/trace/span.ts
+++ b/src/trace/span.ts
@@ -120,10 +120,18 @@ export interface Span {
   isRecording(): boolean;
 
   /**
-   * Sets exception as a span event
+   * Sets exception as a span event.
+   *
    * @param exception the exception the only accepted values are string or Error
+   * @param [attributesOrTime] additional attributes to be associated with
+   *     this event. Can also be the time parameter if type is {@type TimeInput}
+   *     and 3rd parameter is undefined.
    * @param [time] the time to set as Span's event time. If not provided,
    *     use the current time.
    */
-  recordException(exception: Exception, time?: TimeInput): void;
+  recordException(
+    exception: Exception,
+    attributesOrTime?: SpanAttributes | TimeInput,
+    time?: TimeInput
+  ): void;
 }
--- a/src/trace/NonRecordingSpan.ts
+++ b/src/trace/NonRecordingSpan.ts
@@ -71,5 +71,9 @@ export class NonRecordingSpan implements Span {
   }
 
   // By default does nothing
-  recordException(_exception: Exception, _time?: TimeInput): void {}
+  recordException(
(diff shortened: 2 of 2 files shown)


src/trace/span.ts
```diff
@@ -120,10 +120,18 @@ export interface Span {
...
+   *
    * @param exception the exception the only accepted values are string or Error
+   * @param [attributesOrTime] additional attributes to be associated with
+   *     this event. Can also be the time parameter if type is {@type TimeInput}
+   *     and 3rd parameter is undefined.
    * @param [time] the time to set as Span's event time. If not provided,
    *     use the current time.
    */
-  recordException(exception: Exception, time?: TimeInput): void;
+  recordException(
```
Flarna, line 132: Shouldn't we use an overload here to be more specific? e.g. like this:
```
recordException(exception: Exception, time?: TimeInput): void;
recordException(exception: Exception, attributes: SpanAttributes, time?: TimeInput): void;
```

vreynolds: I like that. For symmetry, I would change the `addEvent` as well, since they're pretty much identical interfaces.

vreynolds: Updated in new commit

dyladan:
Generally I like the change. Before approving and merging I'm trying to get guidance from the TC on the backport policy. Since this will require us to go to 1.1.x, will we need to backport fixes to 1.0.x in the future.

dyladan:
From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#api-support

> Bug and security fixes will be backported. Additional feature development is NOT RECOMMENDED.

Seems backporting fixes will be required for the duration of the 1.x lifecycle, and it is not recommended to make changes if possible. I think this change is not high enough priority to break this recommendation. A user can simply call `recordException` and then call `addAttributes`.

src/trace/span.ts
```diff
@@ -65,14 +65,21 @@ export interface Span {
...
+   *
+   * @param name the name of the event.
+   * @param [attributes] the attributes that will be added; these are
+   *     associated with this event.
    * @param [startTime] start time of the event.
    */
   addEvent(
     name: string,
-    attributesOrStartTime?: SpanAttributes | TimeInput,
+    attributes?: SpanAttributes,
```
Flarna, line 82:
```suggestion
    attributes: SpanAttributes,
```
No need to make this optional here. The other overload allows already a call with name only.

src/trace/span.ts
```diff
@@ -120,10 +127,25 @@ export interface Span {
...
+   * Sets exception as a span event.
+   *
+   * @param exception the exception the only accepted values are string or Error
+   * @param [attributes] additional attributes to be associated with this event.
+   * @param [time] the time to set as Span's event time. If not provided,
+   *     use the current time.
+   */
+  recordException(
+    exception: Exception,
+    attributes?: SpanAttributes,
```
Flarna, line 148:
```suggestion
    attributes: SpanAttributes,
```

src/trace/span.ts
Flarna, line 82: strictly speaking this is a breaking change for a typescript user as `addEvent(name, timeInput, timeInput)` is no longer allowed now.

vreynolds: Ha, that's an interseting edge case! Closing this PR either way, but thanks for catching.

longility:
> From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#api-support
> 
> > Bug and security fixes will be backported. Additional feature development is NOT RECOMMENDED.
> 
> Seems backporting fixes will be required for the duration of the 1.x lifecycle, and it is not recommended to make changes if possible. I think this change is not high enough priority to break this recommendation. A user can simply call `recordException` and then call `addAttributes`.

I didn't see an `addAttributes` to where the attributes are part of the span event created by `recordException`. Is your comment stating that it would be `setAttributes` and the attributes will be on the span that the span event created by `recordException`? I'm trying to see if I overlooked where this `addAttributes` is.

longility:
> From https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#api-support
> 
> > Bug and security fixes will be backported. Additional feature development is NOT RECOMMENDED.
> 
> Seems backporting fixes will be required for the duration of the 1.x lifecycle, and it is not recommended to make changes if possible. I think this change is not high enough priority to break this recommendation. A user can simply call `recordException` and then call `addAttributes`.

@dyladan I didn't see an `addAttributes` to where the attributes are part of the span event created by `recordException`. Is your comment stating that it would be `setAttributes` and the attributes will be on the span that the span event created by `recordException`? I'm trying to see if I overlooked where this `addAttributes` is.

dyladan:
It was just a typo. I meant `setAttributes`.

Re-reading the issue and my response though has made me realize that the feature request was to add attributes to the event, not the span. `recordException` doesn't return the event or provide any mechanism for this. Now that we have agreed to start developing new features (1.1 release will come soon) I think this should be reconsidered.

longility:
I definitely would like this to be reconsidered! I have multiple use cases that need this.

dyladan:
There is no exact process. At the last couple SIG meetings we had decided to release 1.1 when #147, #142, and #129 were merged which has now happened. Now we're preparing the SDK so that the release can be as close to the same time as possible. If you want to make an argument that 1.1 should be held for this feature, I would suggest you make a feature request issue (or a PR I guess) and tag the JS maintainers team. If it is not in 1.1, it will have to wait for 1.2

PR #101, opened 2021-07-08
docs: Format example for tracer.startActiveSpan

ad-m:
Example in documentation fro `tracer.startActiveSpan` is misformated:

image

That PR is gonna fix that:

image

I have doubts whether the second example - although technically correct - is optimal for promotion and indication in the documentation.

### DIFF
--- a/src/trace/tracer.ts
+++ b/src/trace/tracer.ts
@@ -50,36 +50,37 @@ export interface Tracer {
    * @param fn function called in the context of the span and receives the newly created span as an argument
    * @returns return value of fn
    * @example
-   *   const something = tracer.startActiveSpan('op', span => {
-   *     try {
-   *       do some work
-   *       span.setStatus({code: SpanStatusCode.OK});
-   *       return something;
-   *     } catch (err) {
-   *       span.setStatus({
-   *         code: SpanStatusCode.ERROR,
-   *         message: err.message,
-   *       });
-   *       throw err;
-   *     } finally {
-   *       span.end();
-   *     }
-   *   });
+   *     const something = tracer.startActiveSpan('op', span => {
+   *       try {
+   *         do some work
+   *         span.setStatus({code: SpanStatusCode.OK});
+   *         return something;
+   *       } catch (err) {
+   *         span.setStatus({
+   *           code: SpanStatusCode.ERROR,
+   *           message: err.message,
+   *         });
+   *         throw err;
+   *       } finally {
+   *         span.end();
+   *       }
+   *     });
+   *
    * @example
-   *   const span = tracer.startActiveSpan('op', span => {
-   *     try {
-   *       do some work
-   *       return span;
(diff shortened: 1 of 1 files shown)


naseemkullah (approved):

naseemkullah:
> I have doubts whether the second example - although technically correct - is optimal for promotion and indication in the documentation.

The idea was to highlight that you could return the span if you wanted to continue to manage it outside of the function, but improvements to the examples are always welcome!

vmarchaud (approved):

Rauno56:
> > I have doubts whether the second example - although technically correct - is optimal for promotion and indication in the documentation.
> 
> The idea was to highlight that you could return the span if you wanted to continue to manage it outside of the function, but improvements to the examples are always welcome!

OP didn't mean a conceptual change. Only formatting!
Thanks, @ad-m!

Doesn't have to be in this PR, but I personally appreciate examples that run more than pseudo-code ones. Making minimal changes(console.log'ing instead of "do some work" and throwing etc.) could make this look a little better.

Rauno56 (approved):

ad-m:
> OP didn't mean a conceptual change. Only formatting!

I perceive that PR as only formatting change. 

Despite of that I added side note about second example, but I believe it is out of scope of that PR. 

> The idea was to highlight that you could return the span if you wanted to continue to manage it outside of the function, but improvements to the examples are always welcome!

Thanks for the clarification. This may actually be useful on rare occasions.

naseemkullah:
> > > I have doubts whether the second example - although technically correct - is optimal for promotion and indication in the documentation.
> > 
> > 
> > The idea was to highlight that you could return the span if you wanted to continue to manage it outside of the function, but improvements to the examples are always welcome!
> 
> OP didn't mean a conceptual change. Only formatting!

While their PR targets formatting they also inquired about the second example (conceptually) :) 

> Doesn't have to be in this PR, but I personally appreciate examples that run more than pseudo-code ones. Making minimal changes(console.log'ing instead of "do some work" and throwing etc.) could make this look a little better.

Yep, up for a PR?

johnbley (approved):

src/trace/tracer.ts
```diff
@@ -50,36 +50,37 @@ export interface Tracer {
...
+   *       } catch (err) {
+   *         span.setStatus({
+   *           code: SpanStatusCode.ERROR,
+   *           message: err.message,
+   *         });
+   *         throw err;
+   *       }
+   *     });
+   *     do some more work
+   *     span.end();
```
dyladan, line 83: I would remove this example. When using startActiveSpan I expect the span to be ended most of the time within the callback.

dyladan: Nevermind I see you only want to do formatting here.

dyladan (approved):

ad-m:
Thanks for your cooperation.

PR #111, opened 2021-07-29
chore: Add a script to automate releases

dyladan:
```shell
$ ./scripts/release.sh -h               
release.sh Usage:

ALWAYS RUN FROM THE ROOT OF THE REPOSITORY

Arguments:
        -h | --help
                Display this help message

        -d | --dry
                Perform a dry run

        -f | --force
                Force execution even if there are uncommitted changes in the workspace

        -c | --clean
                Remove all files from the workspace not tracked by git

        -l | --latest
                Apply the NPM dist-tag @latest after publish

        -t | --tag next
                Which NPM dist-tag to apply to the release. Default: next
```

```shell
$ ./scripts/release.sh -d
Dry run
Skipping cleaning directory
npm install
npm compile
npm test
npm publish --tag next
```

```shell
$ ./scripts/release.sh -d -t mytag                    
Dry run
Skipping cleaning directory
npm install
npm compile
npm test
npm publish --tag mytag
```

```shell
$ ./scripts/release.sh -d -t mytag -l 
Dry run
Skipping cleaning directory
npm install
npm compile
npm test
npm publish --tag mytag
npm dist-tag add "@opentelemetry/api@mytag" latest
```

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -19,8 +19,8 @@ while [[ $# -gt 0 ]]; do
       FORCE=YES
       shift
       ;;
-    -c|--no-clean)
-      NOCLEAN=YES
+    -c|--clean)
+      CLEAN=YES
       shift
       ;;
     -l|--no-latest)
@@ -58,8 +58,8 @@ if [ -n "$HELP" ]; then
   echo -e "\t-f | --force"
   echo -e "\t\tForce execution even if there are uncommitted changes in the workspace"
   echo ""
-  echo -e "\t-c | --no-clean"
-  echo -e "\t\tDo not clean the workspace"
+  echo -e "\t-c | --clean"
+  echo -e "\t\tRemove all files from the workspace not tracked by git"
   echo ""
   echo -e "\t-l | --latest"
   echo -e "\t\tApply the NPM dist-tag @latest after publish"
@@ -71,12 +71,12 @@ if [ -n "$HELP" ]; then
 fi
 
 if [ ! -d ".git" ]; then
-  echo "Not in a git repository"
+  echo "Not in a git repository. Please make sure to run this from the root of the OpenTelemetry API repository."
   exit 1
 fi
 
 if changes=$(git status --porcelain) && [ -n "$changes" ] && [ -z "$FORCE" ]; then
-  echo "There are uncommitted changes in the workspace"
+  echo "There are uncommitted changes in the workspace. Please commit or stash the changes or run this command again with the --force option."
   echo $changes
   exit 1
 fi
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -0,0 +1,116 @@
...
+
+# Ensure working directory is clean if user has not specifically disabled cleaning
+if [ -z "$NOCLEAN" ]; then
+  echo "Cleaning workspace"
+  echo "rm -rf node_modules"
+  [ -z "$DRY" ] && rm -rf node_modules
+  echo "rm -f package-lock.json"
+  [ -z "$DRY" ] && rm -f package-lock.json
+  echo "git clean -fdx"
+  [ -z "$DRY" ] && git clean -fdx
```
obecny, line 94: this is quite dangers, you can accidently loose some untracked files or changes, as you can simply either copy paste command or retry the previously used one. 
I would add some dialog to confirm user really wants that as there is no easy ctrl+z afterwards, otherwise I would be sceptical to run it even with dry run

dyladan: You should not be deploying if you have untracked files or changes. The command will even fail and not run if you have untracked changes unless you do `--force`

dyladan: I reversed the flag so it will not clean by default unless you do `--clean`

Rauno56: If it works for your usecase Daniel, I don't want to bloat the implementation, but another option would be to stash untracked files instead of cleaning them.

Rauno56 (approved):

scripts/release.sh
dyladan: I would honestly prefer the user make sure their workspace is clean before even running the command. The -f and -c flags are only there as conveniences. Releasing is a very deliberate process and shouldn't be something anyone runs on a dirty branch without a very good reason.

obecny: ideally we would have exactly the same github action as in contrib repo, not sure how much time consuming is that though. So this way it will always be safe and easy

dyladan: That is the next step. I am going to create a github action which calls this script. The advantage of doing it this way is we can more easily do the release locally but still in an automated fashion if needed (e.g. releasing bugfix backports on a release/x.y branch)

dyladan: @obecny is the reversed -c flag enough to resolve this?

pragmaticivan:
Not sure if that's too late. Might be worth evaluating https://github.com/googleapis/release-please still?

I have a functional example: https://github.com/pragmaticivan/nestjs-otel/blob/main/.github/workflows/release-please.yml

MSNev (approved):

scripts/release.sh
obecny: yeap fine with me

obecny (approved):

dyladan:
Another contributor has made me aware of a release automation project from google called release-please which handles what this PR was trying to solve. Please take a look at #112 and if we prefer it we can close this PR in favor of that one.

PR #112, opened 2021-08-03
chore: add release-please github action

dyladan:
This PR replaces #111 

The release-please github action will create a release PR automatically when we use conventional commits. When the PR is merged, it will be released to NPM.

release-please updates the package.json but does not run the npm `version` script so in order to solve that I have removed the version.ts file and ensured it will be autogenerated before compile but not checked into git.

- Add release-please action
- Remove version.ts
- Add version.ts to gitignore
- Generate version.ts in precompile script


Flarna:
Could we read our version from package.json instead generating a version.ts file?

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,33 @@
...
+          release-type: node
+          package-name: "@opentelemetry/api"
+      # The logic below handles the npm publication:
+      - uses: actions/checkout@v2
+        # these if statements ensure that a publication only occurs when
+        # a new release is created:
+        if: ${{ steps.release.outputs.release_created }}
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
```
Flarna, line 23:
```suggestion
          node-version: 14
```
Or is there a special reason for sticking on 12?

dyladan:
> Could we read our version from package.json instead generating a version.ts file?

This only works for customers who don't use things like webpack. There is no guarantee our package.json will make it to the final running application.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,33 @@
+on:
+  push:
+    branches:
+      - master
```
Flarna, line 4:
```suggestion
      - main
```

.github/workflows/release-please.yml
dyladan: No it was just what was in their example

.github/workflows/release-please.yml
pragmaticivan, line 4: Is that intentional ^ The default branch for that repo is main. If you want to use master just for tests you might need to add a flag to point that branch on the release please action step

pragmaticivan: In fact you can also "test it" by pointing to the branch in this PR

dyladan: oops forgot to push :)

Flarna (approved):

Rauno56 (approved):

MSNev (approved):

obecny (approved):
lgtm

PR #117, opened 2021-08-13
docs(context): document basic context operation

dyladan:


docs/context.md
```diff
@@ -1,5 +1,99 @@
 # Context
 
-TODO
+In order for OpenTelemetry to work, it must store and propagate important telemetry data.
+For example, when a request is received and a span is started it must be available to a component which creates its child span.
+To solve this problem, OpenTelemetry stores the span in the Context.
+This document describes the OpenTelemetry context API and how it is used.
```
svrnm, line 6: This document describes the OpenTelemetry context API for JavaScript and how it is used.

docs/context.md
```diff
@@ -1,5 +1,99 @@
...
+
+The user may get, set, or delete context values by calling methods [documented here](https://open-telemetry.github.io/opentelemetry-js-api/interfaces/context.html).
+Because context is immutable, modifying an entry does not modify the context.
+Instead, it creates a new context which contains all of the entries of the previous context with the new entry added.
+
+Keys are created by calling [`api.createContextKey(description)`](https://open-telemetry.github.io/opentelemetry-js-api/modules.html#createcontextkey).
+Certain keys, such as the active span, are predefined to have specific meaning by the OpenTelemetry specification and the OpenTelemetry JS API.
+These predefined context properties can only be accessed through use of APIs such as [`api.trace.getSpan()`](https://open-telemetry.github.io/opentelemetry-js-api/classes/traceapi.html#getspan).
+
+```typescript
```
svrnm, line 31: In this example I miss the require statements, it would be good if I could copy&paste the example into a file and try it out from there (if this references an example file, a link to it would be an alternative)

docs/context.md
```diff
@@ -1,5 +1,99 @@
...
+// Creates a new context with boolKey set to true
+const ctx2 = ctx.setValue(boolKey, true);
+
+console.log(ctx.getValue(boolKey)) //? undefined
+console.log(ctx2.getValue(boolKey)) //? true
+```
+
+### Get Active Context
+
+**IMPORTANT**: This assumes you have configured a Context Manager.
```
svrnm, line 46: Reading through this document standalone, to it is not clear what a "Context Manager" is, a reference to the API or any other place on how to configure one would make this reproducable.

docs/context.md
```diff
@@ -1,5 +1,99 @@
...
+Without one, `api.context.active()` will _ALWAYS_ return the `ROOT_CONTEXT`.
+
+The active context is the context object which is returned when `api.context.active()` is called.
+This is accomplished through the use of mechanisms like [async_hooks](https://nodejs.org/api/async_hooks.html) in node or [zone.js](https://github.com/angular/zone.js/) on the web.
+JavaScript is single-threaded, but it can usually be thought of as similar to thread-local storage.
```
svrnm, line 51: What is 'it' referring to? Not "JavaScript", I guess it referres to the context object?

docs/context.md
```diff
@@ -1,5 +1,99 @@
...
+During execution of the `callback`, the context passed to `with` will be returned by `context.active` unless another context is made active.
+
+```typescript
+const ctx =  api.context.active(); // Returns the active context or the ROOT_CONTEXT if no context is active
+const ctx2 = ctx.setValue(boolKey, true); // does not modify ctx
+
+console.log(ctx.getValue(boolKey)) //? undefined
+console.log(ctx2.getValue(boolKey)) //? true
+
+const ret = api.context.with(ctx2, () => {
```
svrnm, line 81: A suggestion to make it easier to follow this example, instead of a `boolKey` you could use a string or an int that is set to "Context 2", "Context 3" or 2 and 3, so I don't have to go back and check what was false and true.

dyladan: Done

docs/context.md
dyladan: I removed the line as it is most likely just confusing.

docs/context.md
dyladan: Added a section for context manager near the top PTAL

docs/context.md
dyladan: Added require statements to each example so they are copy-paste

docs/context.md
```diff
@@ -1,5 +1,124 @@
...
+import { AsyncHooksContextManager } from "@opentelemetry/context-async-hooks";
+
+const contextManager = new AsyncHooksContextManager();
+context.Manager.enable();
+api.context.setGlobalContextManager(contextManager);
+```
+
+## Basic Operations
+
+The context is an immutable map from context keys to any value.
```
MSNev, line 35: technical nit: It's not ___really___ immutable if you can ```set``` stuff to it 😄, suggestion drop ```immutable``` or describe it a bit more.

docs/context.md
```diff
@@ -1,5 +1,124 @@
...
+
+### Context Modification
+
+The user may get, set, or delete context values by calling methods [documented here](https://open-telemetry.github.io/opentelemetry-js-api/interfaces/context.html).
+Because context is immutable, modifying an entry does not modify the context.
```
MSNev, line 40: Possible rewording suggestions

The context is a container that holds the current keyed (named) state (map) of objects, to ensure that the changes to the current state do not corrupt any previous state the API enforces immutability by returning a ```new``` context instance for each modification operation (```setValue```, ```deleteValue```) thus keeping the previous state unchanged.

For example ``` api.context.setValue(newKey, newValue) ``` does NOT add or change the ```newKey``` references within the ```api.context``` but rather returns a new ```context``` instance which contains all of the previous state with the modification applied (set/delete) so you need to capture the new returned context ```var myContext = api.context.setValue(newKey, newValue);``` and use it rather than the global ```api.context```.

MSNev: I have the wrong references above, but hopefully you get the idea... basically describing what is meant by ```immutable``` even though the api includes (apparent) mutable operations.

docs/context.md
```diff
@@ -1,5 +1,124 @@
...
+const myKey = api.createContextKey("Key to store a value");
+
+// Creates a new context with myKey set to "my value"
+const ctx2 = ctx.setValue(myKey, "my value");
+
+console.log(ctx.getValue(myKey)) //? undefined
+console.log(ctx2.getValue(myKey)) //? "my value"
+```
+
+### Get Active Context
```
MSNev, line 62: Maybe move this above the Context Modification for a better flow and context for when someone is reading in isolation.

docs/context.md
vmarchaud: I don't think we should drop it, the API is meant to be immutable even though we don't forbid using API like `Object.freeze` (we could though)

docs/context.md
vmarchaud: i think its worth to add `async_context` link since we use it from node 14.x and upwards (https://nodejs.org/api/async_context.html)

docs/context.md
dyladan: It is in fact immutable since it can't be mutated. When you `set` a new context is created which contains all values from the previous context.

vmarchaud: I think he meant that you can theorically access the context map (inside a context manager) and update it as you want, nothing is preventing you you to do it.

MSNev (approved):

docs/context.md
dyladan: It is not possible using officially supported APIs. You can always find some way to break it. I think it's important to the understanding and proper use of the API to treat it as immutable.

vmarchaud (approved):

PR #123, opened 2021-09-24
feat(context): add attach and detach methods to context API

dyladan:
This adds attach/detach methods to context API as they are specified in [optional global operations](https://github.com/open-telemetry/opentelemetry-specification/blob/b46bcab5fb709381f1fd52096a19541370c7d1b3/specification/context/context.md#optional-global-operations).

Non-breaking for users as this is an additive change.

Breaking for SDK implementers. SDKs must implement the attach/detach methods.


vmarchaud (approved):

src/api/context.ts
```diff
@@ -87,6 +87,27 @@ export class ContextAPI {
...
+   * @param context context to make active in the current asynchronous execution
+   * @returns a restore key
+   */
+  public attach(context: Context): symbol {
+    return this._getContextManager().attach(context);
+  }
+
+  /**
+   * Restore the context which was active when attach was called using the restore
+   * token returned by attach.
```
Flarna, line 103: How is the behavior if the context to detach is not active at this time? e.g. one more call to `attach()` was made or in the middle of a `context.with()`

Flarna:
Have you verified that we can implement these APIs with the existing context managers (AsyncHooks, zone.js)?
If not this would be quite problematic.

src/api/context.ts
```diff
@@ -87,6 +87,27 @@ export class ContextAPI {
     return this._getContextManager().bind(context, target);
   }
 
+  /**
+   * Make a context active in the current execution. Returns a unique restore
+   * key which must be used with detach to restore the previous context.
+   *
```
Flarna, line 93: I recommend to add some warnings here when to use the function similar as for [`asyncLocalStorage.enterWith`](https://nodejs.org/dist/latest-v16.x/docs/api/async_context.html#async_context_asynclocalstorage_enterwith_store)

It happens very easy that an `attach` is done but `detach` is missing for a code patch (only an else, exception, one case in a switch,...). This results in a context mess which is propagated and possible side effects in quite unrelated code areas. Such issues are terrible hard to debug.

src/api/context.ts
vmarchaud: I understood that the expected behavior is to return the point of context before the original attach ? That would require to store a context stack

vmarchaud:
> Have you verified that we can implement these APIs with the existing context managers (AsyncHooks, zone.js)?
If not this would be quite problematic.

Looks doable to me for async hooks at least

Flarna:
I think we should work on prototypes for async hooks and zone.js before adding this to API. Otherwise the risk is too high to loose support for a major part.

src/api/context.ts
Flarna: We should clears specify this. e.g. sequences like:
* attach(a) => attach(b) => detach(a) => detach(b): do we want `a` restored here?
* with(a, () => attach(b)) => detach(b): do we want `a` restored here? What is restored at the end of the `with` cb?
* ...

So it might be something more complex then a stack. Till now it was an implicit stack as an inner `with()` always exited before and other `with()`.

dyladan:
> Have you verified that we can implement these APIs with the existing context managers (AsyncHooks, zone.js)? If not this would be quite problematic.

Yes I have implemented prototypes in the async hooks and async local storage context managers. It was actually much simpler than I expected. I have not tried in the zone context manager yet.


> I think we should work on prototypes for async hooks and zone.js before adding this to API. Otherwise the risk is too high to loose support for a major part.

I also think that's a good idea, but I'm not sure how to expose it for use without adding it to the API. If it is never used, then we will never learn anything from it. Maybe we should add a prefix to the API methods like `api.context.experimental.attach`?

Flarna:
We could move API into core repo and link them via lerna :o).

I think the development of such an API must start go hand in hand with core anyway otherwise core breaks automatically as it uses `"@opentelemetry/api": "^1.0.2"` as dependency. Adding a new mandatory API would break CI once API is published (could be avoided by pinning API dependency in core).

I see two variants:
* merge API and publish a canary version to NPM and use it in core. Once core is done do an API release.
* create API proposal here. Once it settled implement in core and merge to master (maybe even publish to NPM is needed). Once this is done proceed with API.


> Yes I have implemented prototypes in the async hooks and async local storage context managers

Is the semantic of `asyncLocalStorage.exit` and `asyncLocalStorage.enterWith` the same as that one specified here (e.g. in case of nested calls, "wrong" enter/exit sequence,...)?



> I have not tried in the zone context manager yet.

I did a fast look into the API but found nothing like `enterWith`/`exit`.

src/context/types.ts
```diff
@@ -67,6 +67,22 @@ export interface ContextManager {
    */
   bind<T>(context: Context, target: T): T;
 
+  /**
+   * Make a context active in the current execution
+   *
+   * @param context context to make active in the current asynchronous execution
```
blumamir, line 73: Can we elaborate a bit about the `asynchronous execution` meaning?

For example:
### After Await
```js
const contextKey = Symbol('example-context-key');

const g = async () => {
    await new Promise(resolve => setTimeout(resolve, 10));
}

const f = async () => {
    context.attach(context.active().setValue(contextKey, 'example context value'));
    await g();
    console.log(`should the context be attached after await? ${context.active().getValue(contextKey)} === 'example-context-key'?`);
}
```
I would expect that it should

### Sync Function Returns
```js
const contextKey = Symbol('example-context-key');

const g = () => {
    context.attach(context.active().setValue(contextKey, 'example context value'));
}

const f = async () => {
    g();
    console.log(`should the context set in sync function g be active in code after that function returns? ${context.active().getValue(contextKey)} === 'example-context-key'?`);
}
```
I would expect that it should, even thought it makes it non trivial IMO to detach a context in this case since it was set inside a function which already returned.

### Async Function Returns and Awaited
```js
const contextKey = Symbol('example-context-key');

const g = async () => {
    context.attach(context.active().setValue(contextKey, 'example context value'));
}

const f = async () => {
    await g();
    console.log(`should the context set in async function g be active in code after that function is await? ${context.active().getValue(contextKey)} === 'example-context-key'?`);
}
```
I expect that it should.

### context.with Behaviour
```js
const contextKey = Symbol('example-context-key');

const g = () => {
    context.attach(context.active().setValue(contextKey, 'example context value'));
}

const f = () => {
    context.with(context.active().setValue(Symbol('symbol-for-context-with'), 'value'), g);
    console.log(`should the context set inside function g executed with 'context.with' be active after the with returns? ${context.active().getValue(contextKey)} === 'example-context-key'?`);
}
```
This is similar to the previous example, but now the function is executed with `context.with` which, I suppose will restore the original context that was set on `f()`. which means that the context attached in `g()` will probably not active after the `with` returns. 

In the async_hooks mechanisim, the code executed after g is the same "asynchronous execution" context, but the otel context manager might change the behaviour in this case which can lead to confusion IMO.

src/context/types.ts
```diff
@@ -67,6 +67,22 @@ export interface ContextManager {
...
+   */
+  attach(context: Context): symbol;
+
+  /**
+   * Restore the context which was active when attach was called using the restore
+   * token returned by attach.
+   *
+   * @param token the restore token returned by attach
+   */
+  detach(token: symbol): void;
```
blumamir, line 84: Just to make sure I understand - `detach` is optional, right? if the user choose too, he can set a context and then forget about it and it will remain active for the entire `asynchronous execution` and die when it terminates.

src/api/context.ts
```diff
@@ -87,6 +87,27 @@ export class ContextAPI {
     return this._getContextManager().bind(context, target);
   }
 
+  /**
+   * Make a context active in the current execution. Returns a unique restore
```
blumamir, line 91: The term here is "current execution", but below we use the term "asynchronous execution".
I would suggest to stick with one term so it is less confusing and more consistent for users.

src/context/NoopContextManager.ts
```diff
@@ -35,6 +35,12 @@ export class NoopContextManager implements types.ContextManager {
     return target;
   }
 
+  public attach(_context: types.Context): symbol {
+    return Symbol('NoopContextManager context restore key');
```
blumamir, line 39: Maybe we can add "OpenTelemetry" similar to:
[`const SPAN_KEY = createContextKey('OpenTelemetry Context Key SPAN');`](https://github.com/open-telemetry/opentelemetry-js-api/blob/main/src/trace/context-utils.ts#L26).

In my experience, these value can be printed for debug and then it really helps that they explain the context in the name.

blumamir:
To summarize my biggest concern with the current direction is that, to my understanding, the attached key will only be active for the duration of the top most wrapping `context.with` call. When the `context.with` callback terminate, the previous context will be restored and will not contain the attached key.
That might come as a surprise to users and be counterintuitive.

I think the expectation of the user is that the context remains attached for any code that "follows" the call. For example - if it was set in a framework's (like fastify) middleware it should also be active on any downstream operation like other middlewares and request handlers. If the key is lost when the middleware returns and the middleware span's context is ended, then it's not solving the need IMO.

dyladan:
I think not calling `detach` after calling `attach` should be considered a bug. You can't expect your context that you `attach`'d to survive the end of a `with`.

You gave the following example:

```typescript
const contextKey = Symbol('example-context-key');

const g = async () => {
    context.attach(context.active().setValue(contextKey, 'example context value'));
}

const f = async () => {
    await g();
    console.log('since g is async, the context attached in g will not be active here');
}
```

I'm not actually sure this is correct. I'll test it with my prototype implementations in the context managers.

Flarna:
A `detach` happens automatically if the current JS tick function ends. So it might be correct to not call detach if you want the context active until the resource vanishes.

There are usecaes when your code runs in a callback of some framework and you want that *everything* - including the parts done by the framework after your handler returns - are linked to your context.
In a lot cases you can archive this by patching something deep in the framework which covers all these parts. e.g the HTTP instrumentation wraps the `request` handler by patching `createServer`.
A less intrusive/patchless variant would be to just prepend a `request` listener and do an attach there.

But this comes with high risk. The framework in use may do quite some work after the user part returned. And the user who didn't a detach should know what happens there and be sure that the context should be really active in these code parts.

stefee (approved):

src/context/types.ts
stefee: @blumamir Your concern about setting active context inside an async function behaving different to non-async seems valid, since this is a language/runtime implementation quirk and not part of the spec.

I’m not sure what to suggest here other than to continue with the ALS implementation and make sure to document its quirks if there are some - let’s test properly to make sure we understand how it works. @dyladan seemed uncertain further down in the thread if your assumptions were correct.

For the specific Fastify use case, I think it would make sense for Fastify to provide [diagnostics channels](https://nodejs.org/dist/latest-v16.x/docs/api/diagnostics_channel.html) rather than relying on lifecycle hooks.

weyert:
@open-telemetry/javascript-maintainers How can we progress this PR? Looks useful

stefee:
I think this might be a hard requirement for https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1021 - until then I don't know if there is a workaround for Undici users other than to manually add propagation headers on outgoing requests.

src/context/types.ts
dyladan: from the spec https://opentelemetry.io/docs/reference/specification/context/#attach-context

>Note that every call to this operation should result in a corresponding call to [Detach Context](https://opentelemetry.io/docs/reference/specification/context/#detach-context).

src/api/context.ts
dyladan: Updated the comment with more warnings, uses, and examples. You can run `npm run docs` and open `docs/out/classes/contextapi.html#attach` to see what the docs look like rendered.

src/api/context.ts
```diff
@@ -87,6 +87,27 @@ export class ContextAPI {
...
+   * key which must be used with detach to restore the previous context.
+   *
+   * @param context context to make active in the current asynchronous execution
+   * @returns a restore key
```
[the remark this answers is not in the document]
dyladan: This is done automatically by release automation now

src/api/context.ts
dyladan: IMO it should act like a stack and detaching out of order would simply result in loss of whatever part of the stack was made active in the meantime:

```typescript
// root is active at start
akey = attach(a) // a is active
bkey = attach (b) // b is active
detach(akey) // root is active, context a can never be restored
detach(bkey) // root is still active
```

src/api/context.ts
```diff
@@ -87,6 +87,27 @@ export class ContextAPI {
...
+  }
+
+  /**
+   * Restore the context which was active when attach was called using the restore
+   * token returned by attach.
+   *
+   * @param token the restore token returned by attach
+   */
+  public detach(token: symbol): void {
+    return this._getContextManager().detach(token);
```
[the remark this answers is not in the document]
dyladan: The versioning guidelines state that the API can use new SDK methods. The end user should be able to update their SDK.

src/context/types.ts
dyladan: Lot of technical details to unpack.

I agree with @vmarchaud on all points except "Async Function Returns and Awaited".

> ```js
> const contextKey = Symbol('example-context-key');
> 
> const g = async () => {
>     context.attach(context.active().setValue(contextKey, 'example context value'));
> }
> 
> const f = async () => {
>     await g();
>     console.log(`should the context set in async function g be active in code after that function is await? ${context.active().getValue(contextKey)} === 'example-context-key'?`);
> }
> ```

In this example I think the context may leak and would be considered undefined behavior because the user failed to call detach. Every attach should have a corresponding detach as stated in the specification and in the docs on attach.

src/context/types.ts
dyladan: >if the user choose too, he can set a context and then forget about it and it will remain active for the entire asynchronous execution and die when it terminates.

spec doesn't allow this, but yes this is how it would work

legendecas (commented):
Not sure if it has been pointed out already. One concern I have about the `attach` and `detach` API is that [`AsyncLocalStorage.enterWith`](https://nodejs.org/api/async_context.html#asynclocalstorageenterwithstore)/`AsyncLocalStorage.exit` are still experimental, while at the moment `AsyncLocalStorage` is one of the significant and stable context manager implementations of the context manager API. This can degrade the context manager implementation to depend on experimental APIs.

But I suppose that `AsyncLocalStorageContextManager` already utilized experimental APIs like `AsyncLocalStorageContextManager.disable` [[1]](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-context-async-hooks/src/AsyncLocalStorageContextManager.ts#L48)  so it can be acceptable?

vmarchaud:
> But I suppose that AsyncLocalStorageContextManager already utilized experimental APIs like AsyncLocalStorageContextManager.disable [[1]](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-context-async-hooks/src/AsyncLocalStorageContextManager.ts?rgh-link-date=2022-06-27T02%3A31%3A19Z#L48) so it can be acceptable?

I mean we started using ALC when it was still experiemental so i don't think its an issue by itself. I would even add that `attach`/`detatch` would not be used most of time so even if there are an issue with the underlying implementation, it would not be a problem for the majority of our users ?

dyladan:
> > But I suppose that AsyncLocalStorageContextManager already utilized experimental APIs like AsyncLocalStorageContextManager.disable [[1]](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-context-async-hooks/src/AsyncLocalStorageContextManager.ts?rgh-link-date=2022-06-27T02%3A31%3A19Z#L48) so it can be acceptable?
> 
> I mean we started using ALC when it was still experiemental so i don't think its an issue by itself. I would even add that `attach`/`detatch` would not be used most of time so even if there are an issue with the underlying implementation, it would not be a problem for the majority of our users ?

We can also add a disclaimer to the JS docs that it uses experimental APIs

dyladan:
Added experimental tags to the docs and verified the rendered docs look as expected.

src/context/types.ts
vmarchaud: @dyladan Do you know what happen with ALS in this case ?

vmarchaud:
> Added experimental tags to the docs and verified the rendered docs look as expected

I think that's fine, i mean in the end people that want to use it will do it even if we had some protections so lets just mark it as experimental and move forward, even if someone complain we would just redirect to the fact that it was experimental.
Apart from https://github.com/open-telemetry/opentelemetry-js-api/pull/123#discussion_r732716054, lgtm

src/context/NoopContextManager.ts
```diff
@@ -35,6 +35,12 @@ export class NoopContextManager implements types.ContextManager {
     return target;
   }
 
+  public attach(_context: types.Context): symbol {
+    return Symbol('OpenTelemetry NoopContextManager context restore key');
```
blumamir, line 39: nit: we already create symbol in the repo [with this style](https://github.com/open-telemetry/opentelemetry-js-api/blob/bafcf0d0756e6ecc79fcc75f221b9dd17641eba5/src/internal/global-utils.ts#L27): `opentelemetry.js.api.${major}`.
WDYT about changing this text be consistent, e.g: `opentelemetry.js.context.noop_context_manager.restore_key`?

blumamir (approved):
LGTM

src/api/context.ts
```diff
@@ -87,6 +87,89 @@ export class ContextAPI {
...
+   *   api.context.active()     // returns ctx1
+   *   api.context.attach(ctx2) // make ctx2 active
+   *   api.context.active()     // returns ctx2
+   * }
+   * ```
+   * 
+   * @param context context to make active in the current execution
+   * @returns a restore key
+   */
+  public attach(context: Context): symbol {
```
Flarna, line 156: I think attach should return a function or some object with a detach method instead a symbol. Otherwise the context manager needs a map or similar to keep track of the symbol => context mapping which could easily result in a mem leak as CM can't know when to clean entries from the map (in special if users forget about calling detach).

Flarna: spec says `The API MUST return a value...`. To my understanding in JS a function is a value as closers are allocated and GCed like strings and numbers.

By the way, I recommend to start implementing the new APIs for zone.js and AsyncLocalStorage CM already before this is merged. In the real implementations it's possible to write a set of tests for all the special cases one can think about (good and bad cases).
This eases review and it proves also that API is portable to browser and node.js. I think it's important the both implementations show the same behavior.

vmarchaud: It looks like the pattern that is used by react hooks but i don't have preference, however it seems weird that the spec say that we must have a return value, what other languages are doing ?

> By the way, I recommend to start implementing the new APIs for zone.js and AsyncLocalStorage CM already before this is merged. In the real implementations it's possible to write a set of tests for all the special cases one can think about (good and bad cases).

I believe Daniel said that he has a working POC for both ?

dyladan: I did but i couldn't find it so i'm reimplementing it. Would be a lot easier if API was in the main repo again

src/api/context.ts
```diff
@@ -87,6 +87,89 @@ export class ContextAPI {
...
+   * function func1() {
+   *   api.context.attach(ctx1)
+   * }
+   * 
+   * function func2() {
+   *   api.context.active() // returns ctx2
+   * }
+   * ```
+   * 
+   * @example <caption>Example of incorrect use of context.attach inside a context.with callback</caption>
```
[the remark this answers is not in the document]
dyladan: Added reasoning. PTAL and let me know if it is clear

src/context/NoopContextManager.ts
dyladan: It's a little different because that is using `Symbol.for` but it seems like a reasonable request

dyladan: It's a little different because that is using Symbol.for but it seems like a reasonable request

src/api/context.ts
Flarna: I was thinking of more than a POC - a complete testsuite reused by all relevant context managers (zone.js, ALS, AsyncHooks) which includes also the non "trivial" cases.

I assume here that we aim equal behavior of these managers. As an alternative we might keep parts of the API description open to allow differences in the implementation.

src/api/context.ts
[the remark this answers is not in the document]
Flarna: But how should someone know that code is running within a `.with()` callback? Maybe the intention of this user is to attach the context till the end of the tick so not calling `.detach()` is the intention?

It doesn't fit to the sentence `Note that every call to this operation should result in a corresponding call to {@link detach} in the reverse order.` is actually not describing the requirements.

It should be more like `detach() MUST be called in the same function as attach() without an await or similar inbetween`.

Or this sample is described as `probably incorrect` or `questionable`.

src/api/context.ts
dyladan: Yes I would hope for the same behavior between all context managers. Having a single test suite which runs against all context managers seems like a great idea. My assertion that all of this would be easier if API was back in main repo stands.

Flarna: If we can't get the same behavior we could specify the behavior on wrong/unexpected/unsupported/... use as undefined. A concrete ContextManager could then specify in more detail what happens in at least some of these cases.

Flarna: I would expect that whatever is written in API is mandatory for implementations. We don't specify in API what/how propagation happen but we specify that after activate/with a get returns this context (sync calls).

But well, from that point of view `NoopContextManager` is not conform to API as get returns always `undefined`.

vmarchaud: > I would expect that whatever is written in API is mandatory for implementations

We will not have in the short term (2-3year imo) a fully working context manager in the browser so saying that the API requires something that we know is not possible seems counter productive even though i agree that in theory i would like that the API enforces behavior.

arminru:
@dyladan Can this one be closed as per https://github.com/open-telemetry/community/issues/1290?

PR #127, opened 2021-10-22
chore: align linter rules with repo open-telemetry/opentelemetry-js

legendecas:
The current linter rule is not exactly the same as the one of open-telemetry/opentelemetry-js, which makes moving codes between them require unnecessary re-formats.


vmarchaud:
cc @open-telemetry/javascript-approvers could someone review this one ? its a straightforward change and its blocking other PRs

Rauno56 (approved):

PR #129, opened 2021-10-29
feat(trace): add optional schema url to TracerProvider.getTracer

### LINKED ISSUE #125: Support spec v1.4.0 schema url in tracer
Spec [v1.4.0](https://github.com/open-telemetry/opentelemetry-specification/blob/main/CHANGELOG.md#v140-2021-06-07) added a new optional parameter to `TracerProvider.getTracer`, ([link](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer)).

However, in a similar situation, implementations like https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-sdk-metrics-base/src/MeterProvider.ts#L48 may already extending the `MeterProvider.getMeter` with positional parameters, so adding new positional parameter `schemaUrl` in `@opentelemetry/api-metrics` would conflict with those implementations. (Related: https://github.com/open-telemetry/opentelemetry-js/pull/2529)

The good news is that `BasicTracerProvider.getTracer` in https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-sdk-trace-base/src/BasicTracerProvider.ts#L93 didn't extend the method with positional parameters.  So we can still extend the method with optional positional parameters:
```typescript
interface TracerProvider {
  getTracer(name: string, version?: string, schemaUrl?: string): Tracer;
}
```

Still, the approach above leaves us a dilemma when a new optional parameter is added in the future. For better extensibility and to keep compatibility, we can override the function with a variable second parameter type like the following example.

```typescript
interface TracerProvider {
  getTracer(name: string, version?: string): Tracer;
  getTracer(name: string, options?: TracerOptions): Tracer;
}
```

legendecas:
According to the conclusion at https://github.com/open-telemetry/opentelemetry-js/pull/2529#discussion_r737629152, the new optional parameter of `TracerProvider.getTracer` should be put into an option bag to prevent from breaking existing SDK implementations. In this way, we can also reserve the possibility for future extensions.

Fixes https://github.com/open-telemetry/opentelemetry-js-api/issues/125
Fixes https://github.com/open-telemetry/opentelemetry-js/issues/2535

### DIFF
--- a/src/trace/ProxyTracerProvider.ts
+++ b/src/trace/ProxyTracerProvider.ts
@@ -18,6 +18,7 @@ import { Tracer } from './tracer';
 import { TracerProvider } from './tracer_provider';
 import { ProxyTracer } from './ProxyTracer';
 import { NoopTracerProvider } from './NoopTracerProvider';
+import { TracerOptions } from './tracer_options';
 
 const NOOP_TRACER_PROVIDER = new NoopTracerProvider();
 
@@ -35,10 +36,10 @@ export class ProxyTracerProvider implements TracerProvider {
   /**
    * Get a {@link ProxyTracer}
    */
-  getTracer(name: string, version?: string): Tracer {
+  getTracer(name: string, version?: string, options?: TracerOptions): Tracer {
     return (
-      this.getDelegateTracer(name, version) ??
-      new ProxyTracer(this, name, version)
+      this.getDelegateTracer(name, version, options) ??
+      new ProxyTracer(this, name, version, options?.schemaUrl)
     );
   }
 
@@ -53,7 +54,11 @@ export class ProxyTracerProvider implements TracerProvider {
     this._delegate = delegate;
   }
 
-  getDelegateTracer(name: string, version?: string): Tracer | undefined {
-    return this._delegate?.getTracer(name, version);
+  getDelegateTracer(
+    name: string,
+    version?: string,
+    options?: TracerOptions
+  ): Tracer | undefined {
+    return this._delegate?.getTracer(name, version, options);
   }
 }
--- a/src/trace/ProxyTracerProvider.ts
+++ b/src/trace/ProxyTracerProvider.ts
@@ -39,7 +39,7 @@ export class ProxyTracerProvider implements TracerProvider {
   getTracer(name: string, version?: string, options?: TracerOptions): Tracer {
     return (
       this.getDelegateTracer(name, version, options) ??
-      new ProxyTracer(this, name, version, options?.schemaUrl)
+      new ProxyTracer(this, name, version, options)
     );
   }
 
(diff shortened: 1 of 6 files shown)


src/trace/ProxyTracerProvider.ts
```diff
@@ -35,10 +36,10 @@ export class ProxyTracerProvider implements TracerProvider {
   /**
    * Get a {@link ProxyTracer}
    */
-  getTracer(name: string, version?: string): Tracer {
+  getTracer(name: string, version?: string, options?: TracerOptions): Tracer {
     return (
-      this.getDelegateTracer(name, version) ??
-      new ProxyTracer(this, name, version)
+      this.getDelegateTracer(name, version, options) ??
+      new ProxyTracer(this, name, version, options?.schemaUrl)
```
obecny, line 42: For consistency I would keep options 
```suggestion
      new ProxyTracer(this, name, version, options)
```

obecny (approved):

vmarchaud (approved):

vmarchaud:
ping @dyladan :)

MSNev (approved):

rauno56 (approved):

src/trace/NoopTracerProvider.ts
```diff
@@ -25,7 +26,11 @@ import { TracerProvider } from './tracer_provider';
  * All operations are no-op.
  */
 export class NoopTracerProvider implements TracerProvider {
-  getTracer(_name?: string, _version?: string): Tracer {
+  getTracer(
+    _name?: string,
+    _version?: string,
+    _options?: TracerOptions
```
dyladan, line 32: I'm actually wondering now if we should use an options object or just put the schema url as a parameter. I know we've already talked about this, but the more I think about it the more I think we should use a param instead of options object. schema url is an identity parameter for the tracer, not really an option/configuration right?

Maybe we should talk about this at SIG

dyladan (commented):
Everything looks good % the thread comment i'd like to discuss tomorrow at SIG

src/trace/NoopTracerProvider.ts
dyladan: On thinking about this more I think it's probably better to leave decisions already made alone

dyladan (approved):

dyladan:
@legendecas feel free to merge this when you're ready

blumamir (approved):

src/trace/tracer_provider.ts
```diff
@@ -31,5 +32,5 @@ export interface TracerProvider {
    * @param version The version of the tracer or instrumentation library.
```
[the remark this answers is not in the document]
legendecas: Updated :)

PR #142, opened 2021-12-16
feat: define common attributes type

dyladan:
As discussed in SIG, this creates a common definition for attributes which can be used by all signals.

Span attributes is marked as deprecated in favor of the new common attributes definition.

### DIFF
--- a/src/common/Attributes.ts
+++ b/src/common/Attributes.ts
@@ -14,14 +14,22 @@
  * limitations under the License.
  */
 
-import { Attributes, AttributeValue } from '../common/Attributes';
+/**
+ * Attributes is a map from string to attribute values.
+ */
+export interface Attributes {
+    [attributeKey: string]: AttributeValue | undefined;
+}
 
 /**
- * @deprecated please use Attributes
+ * Attribute values may be any non-nullish primitive value except an object.
+ *
+ * null or undefined attribute values are invalid and will result in undefined behavior.
  */
-export type SpanAttributes = Attributes;
-
-/**
- * @deprecated please use AttributeValue
- */
-export type SpanAttributeValue = AttributeValue;
+export type AttributeValue =
+    | string
+    | number
+    | boolean
+    | Array<null | undefined | string>
+    | Array<null | undefined | number>
+    | Array<null | undefined | boolean>;
--- a/src/index.ts
+++ b/src/index.ts
@@ -18,6 +18,7 @@ export * from './baggage/types';
 export { baggageEntryMetadataFromString } from './baggage/utils';
 export * from './common/Exception';
 export * from './common/Time';
+export * from './common/Attributes';
 export * from './diag';
 export * from './propagation/TextMapPropagator';
 export * from './trace/attributes';


src/common/Attributes.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * null or undefined attribute values are invalid and will result in undefined behavior.
+ */
+export type AttributeValue =
+    | string
+    | number
+    | boolean
+    | Array<null | undefined | string>
+    | Array<null | undefined | number>
+    | Array<null | undefined | boolean>;
```
Flarna, line 35: add a newline at end of file

Flarna (approved):

src/common/Attributes.ts
dyladan: There already is a newline. Github shows a red icon if the newline is missing

mwear (approved):
LGTM. Since we are marking `SpanAttributes` as deprecated, should we have a follow up task to replace usage with `Attributes`?

pirgeo (approved):

src/common/Attributes.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Attribute values may be any non-nullish primitive value except an object.
+ *
+ * null or undefined attribute values are invalid and will result in undefined behavior.
+ */
+export type AttributeValue =
+    | string
+    | number
+    | boolean
+    | Array<null | undefined | string>
```
legendecas, line 33: The spec says "`null` values SHOULD NOT be allowed in arrays".

> https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute

legendecas (changes requested):
Since there are already 3 approvals, mark the PR as changes requested to prevent accidental merging until we solve the questions.

src/common/Attributes.ts
Flarna: Removing `null` here would be a breaking change.
As it is just a SHOULD an not a MUST statement I would consider it as ok because JS is one of the `languages that do not have appropriate compile-time type checking`).

legendecas: JS can always disrespect these typescript notations. Here we are writing TypeScript which has a compile-time type checking, no?

Flarna: Well, the repos are named opentelemetry-js not opentelemetry-ts so not sure if JavaScrit or TypeScript is the main target.

Not sure then if changes in type definition like this should be considered as breaking at all. I was assuming yes as there are a lot users out there which also use typescript and for them such a change could break their code.

Also the current Trace SDK handles `null` entries in arrays as far as I remember and it has also reached GA.

legendecas (approved):

src/common/Attributes.ts
dyladan: I would consider any change that breaks the types to be a breaking change as a project which uses our project may work one day and fail to compile the next if they update. I took the types exactly as they were before unchanged for this reason.

blumamir (approved):

vmarchaud (approved):

PR #146, opened 2021-12-27
chore: update actions/checkout to v2

### LINKED ISSUE #144: [Security] Workflow test.yaml is using vulnerable action actions/checkout
The workflow test.yaml is referencing action actions/checkout using references v1. However this reference is missing the commit [a6747255bd19d7a757dbdda8c654a9f84db19839](https://github.com/actions/checkout/commits/a6747255bd19d7a757dbdda8c654a9f84db19839) which may contain fix to the some vulnerability. 
The vulnerability fix that is missing by actions version could be related to: 
(1) CVE fix 
(2) upgrade of vulnerable dependency 
(3) fix to secret leak and others. 
Please consider to update the reference to the action.

dyladan:
Fixes #144

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -15,8 +15,14 @@ jobs:
     container:
       image: ${{ matrix.container }}
     steps:
+      - name: Permission Setup
+        run: chmod -R 777 /github /__w
+
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v2
+
+      - name: Permission Setup
+        run: chmod -R 777 /github /__w
 
       - name: Install Dependencies
         run: npm install
@@ -36,10 +42,11 @@ jobs:
     container:
       image: circleci/node:12-browsers
     steps:
-      - name: Checkout
-        uses: actions/checkout@v1
       - name: Permission Setup
         run: sudo chmod -R 777 /github /__w
+
+      - name: Checkout
+        uses: actions/checkout@v2
 
       - name: Install Dependencies
         run: npm install
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -10,19 +10,16 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        container: ["node:8", "node:10", "node:12", "node:14", "node:16"]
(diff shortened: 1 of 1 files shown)


naseemkullah (approved):

vmarchaud:
Is this ready to review since there are debug `ls` in the workflows ? I believe not but i'm not sure

dyladan:
@vmarchaud I opened it as "ready" initially because it seemed so obvious, but i can't get the test to pass on browser or node 16. I'm getting weird permission errors even though the result of `id` shows that we are running as root.

dyladan:
@vmarchaud it is now ready

.github/workflows/test.yaml
```diff
@@ -36,11 +39,12 @@ jobs:
     container:
       image: circleci/node:12-browsers
     steps:
-      - name: Checkout
-        uses: actions/checkout@v1
       - name: Permission Setup
         run: sudo chmod -R 777 /github /__w
 
+      - name: Checkout
```
vmarchaud, line 45: shouldn't this job requires to install git too ?

dyladan: Actually this is just a leftover from the testing. I removed it.

.github/workflows/test.yaml
```diff
@@ -15,8 +15,14 @@ jobs:
     container:
       image: ${{ matrix.container }}
     steps:
+      - name: Permission Setup
+        run: chmod -R 777 /github /__w
```
legendecas, line 19: Just out of curiosity: why do we need this?

legendecas (approved):

.github/workflows/test.yaml
dyladan: Trust me I am also curious. I was having all sorts of problems with the checkout failing without it, but only on the node 16 container. Thats why there was so much thrash on this PR.

vmarchaud: I'm not really confortable merging this, shoudn't this kind of issue be addressed upstream in the actions itself ?

dyladan: I updated it to use setup-node instead of the official node docker container and that seems to have cleared it up.

vmarchaud (approved):

PR #147, opened 2021-12-31
Add TraceState implementation to API

srikanthccv:
Follow up based on 29/12 call, `TraceState` implementation should be part of API for various reasons (Should be able to use in creating in custom propagator only by taking dependency on `@opentelemetry/api` , library instrumentation, specification). There is already an interface `TraceState` exported so introducing the factory function `createTraceState` that returns new `TraceState` implementation.

It is worth noting that there are other components that are supposed to be in api/sdk but are part of `opentelemetry-core` package but `TraceState` is something I needed at the moment.


src/trace/internal/tracestate-impl.ts
```diff
@@ -0,0 +1,110 @@
+/*
```
legendecas, line 1: Do we have a style guide for file names? From what I know, we are naming the files with the class it exporting, in a case-sensitive way.

lonewolf3739: I am not not sure if it exists, couldn't find one. There is an interface `TraceState` exported currently, It doesn't make sense to have the interface since any implementation should adhere to w3c spec. Made changes to honour the backward compatibility and still have the ability to create a tracestate from string.

dyladan: There is no guidance right now. Some files have CamelCase and some files have kebab-case. Usually if the file is a class it matches the class name, but not always. Even within the same package we are not very consistent.

legendecas: I think it would be better if we put guidance on this. Like, the default file systems on macOS and windows are not case sensitive, it could be erroneous if those files are not in a consistent naming pattern. TypeScript also provides checks on them: https://www.typescriptlang.org/tsconfig#forceConsistentCasingInFileNames. 

Anyway, this is not a blocking issue on this PR. :)

dyladan: That typescript option only causes an issue if your import statement doesn't match the case of the file you're importing. It doesn't make different files have consistent casing.

I agree we should have guidance. I'll create an issue in the main repo to add it to the style guide

dyladan:
Is this the exact implementation from core or have you made any notable changes?

>It is worth noting that there are other components that are supposed to be in api/sdk but are part of opentelemetry-core package but TraceState is something I needed at the moment.

Can you create issues for those components at least so we have something to track?

lonewolf3739:
>Is this the exact implementation from core or have you made any notable changes?

The only new change introduced in this PR `createTraceState` since there is already exported interface with `TraceState` name. Everything else is copied over from core package.

>Can you create issues for those components at least so we have something to track?

Sure, will do that.

Flarna:
Are there already plans how to handle this move to API in `@opentelemetry/core` and other related packages?

* `@opentelemetry/core` is GA so removing this class there would be breaking. Redirecting won't work as core exposes the constructor not the factory method which is not accessible in API.
* There are some propagators which use the `TraceState` from core which could move to the API variant but this would require to bump the min API version there.
* How to ensure compatibility between old/new implementation? I think it's likely to end up having both are in used within a single app.

lonewolf3739:
@dyladan 's recommendation was to move this initially and then update wherever it is used. Encourage people to use the api method and issue depreciation notice to core component. But we will continue to keep maintaining the same changes in two places until the next major version.

vmarchaud:
Even though the core is GA we could just make a major bump, its supposed to be a internal package for instrumentations/sdks not something public facing right ?
In the order:
- publish new api (as a minor ?)
- update all usage from all packages to use the api instead of core
- major bump core
- update new version for all packages

Flarna:
I think the first two steps are fine for now. Maybe add a doc deprecation of `TraceState` in core.

There are a few other issues open to move more from core to API. Maybe core gets anyway empty at some point in time...

dyladan:
Right now having all stable SDK packages have the same version makes it a lot easier for end users. Also, our main repo is not really well set up to have packages independently versioned. I think publishing a minor API, updating usage, and adding a deprecation notice in the core package is sufficient for now.

Before merging this, are there any unreleased PATCH level changes that we should release first?

src/trace/internal/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { TraceStateImpl } from './tracestate-impl';
+
+
+export function createTraceState(rawTraceState?: string): TraceStateImpl {
```
dyladan, line 20: Returning the interface only will prevent versioning issues if an internal/private property of the class is changed in the future. Users should only use the documented interface anyway.

```suggestion
export function createTraceState(rawTraceState?: string): TraceState {
```

vmarchaud (approved):

src/trace/internal/tracestate-impl.ts
```diff
@@ -0,0 +1,110 @@
...
+    this._internalState = rawTraceState
+      .split(LIST_MEMBERS_SEPARATOR)
+      .reverse() // Store in reverse so new keys (.set(...)) will be placed at the beginning
+      .reduce((agg: Map<string, string>, part: string) => {
+        const listMember = part.trim(); // Optional Whitespace (OWS) handling
+        const i = listMember.indexOf(LIST_MEMBER_KEY_VALUE_SPLITTER);
+        if (i !== -1) {
+          const key = listMember.slice(0, i);
+          const value = listMember.slice(i + 1, part.length);
+          if (validateKey(key) && validateValue(value)) {
```
blumamir, line 82: not directly related to this PR, but would love to hear opinions:
looks like currently, we validate keys \ values and other stuff only when constructing a new `TraceState` from string.
Does it also make sense to do some validations when we `set` a new key-value pair on the existing TraceState?

[Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-handling.md#key) does not require it explicitly, but it makes sense to not serialize keys and values which we know will fail to parse down-stream

src/trace/internal/validators.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/**
+ * Value is opaque string up to 256 characters printable ASCII RFC0020
+ * characters (i.e., the range 0x20 to 0x7E) except comma , and =.
+ */
+export function validateValue(value: string): boolean {
+  return (
+    VALID_VALUE_BASE_REGEX.test(value) &&
+    !INVALID_VALUE_COMMA_EQUAL_REGEX.test(value)
```
blumamir, line 43: Does all these values refer specifically to trace state?
There is nothing in the code that references that `VALID_KEY` is actually `VALID_TRACE_STATE_KEY`, except for tracing the usages in other files.
I guess it can be confusing to future people who will read this code.

WDYT about placing these in a file that has `trace_state` in the name, or adding `TraceState` to the variables to explain the context? (like the tests for this file which are called `tracestate-validators.test.ts`)

blumamir (approved):

src/trace/internal/tracestate-impl.ts
lonewolf3739: Agree. I noticed this while looking at the test cases. I didn't want to change any code in this PR but only copy over the files from another package.

src/trace/internal/validators.ts
lonewolf3739: Yes, this is based of this W3C spec https://www.w3.org/TR/trace-context/#key

blumamir: So what do you think about changing the file name to `tracestate-validators.ts`, so it's easier to have a context for these values when reading the code?
This will also align with the test file name: `test/trace/tracestate-validators.test.ts`

dyladan (approved):

legendecas (approved):

PR #159, opened 2022-04-05
docs(api): update earliest supported node version

svetlanabrennan:
Signed-off-by: Svetlana Brennan <svetlana.svn@gmail.com>

Partially fixes Issue #[2721](https://github.com/open-telemetry/opentelemetry-js/issues/2721)

- See checklist item 1 mentioned [here](https://github.com/open-telemetry/opentelemetry-js/issues/2721#issuecomment-1076570990)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ The methods in this package perform no operations by default. This means they ca
 
 - An application you wish to instrument
 - [OpenTelemetry JS SDK][opentelemetry-js]
-- Node.js >=8.5.0 (14+ is preferred) or an ECMAScript 5+ compatible browser
+- Node.js >=8.12.0 (14+ is preferred) or an ECMAScript 5+ compatible browser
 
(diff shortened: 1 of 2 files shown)


Flarna (approved):

legendecas (approved):

vmarchaud (commented):
can we change this on the API without a major bump ?

EDIT: I understand that the SDK requires 8.12 for correct timestamp but the api do not right ?

Flarna:
I did a quick look and it seems API is not using `perf_hooks` so it will even work with node.js 8.0.0.

vmarchaud:
> I did a quick look and it seems API is not using perf_hooks so it will even work with node.js 8.0.0.

API should be platform agnostic anyway so we should keep the support i think

Rauno56:
> I did a quick look and it seems API is not using `perf_hooks` so it will even work with node.js 8.0.0.

That's right. @svetlanabrennan, perhaps we can close this PR then as technically API is not restricted.

Issue #162, opened 2022-06-09
Special case API for getting the current span


cartermp:
Today, the way we document getting the current span is to pass in the active context to `getSpan`:

```js
const currentSpan = opentelemetry.trace.getSpan(opentelemetry.context.active());
```

Conceptually, this is seen as one operation, but it's actually two (get the active context, then get the span from that).

Some feedback I've seen from otel-js users is that it'd be nice if they could have a special case API, perhaps like this:

```js
const currentSpan = opentelemetry.trace.getCurrentSpan();
```

I think it's quite common to want whatever the current span is and then do something with it, but less common to want an arbitrary span from an arbitrary context. My interpretation of the spec is that [this should be allowable](https://opentelemetry.io/docs/reference/specification/trace/api/#context-interaction). If so, the `getCurrentSpan` function would be my initial preference, but really anything that shortens the code and requires a single operation, rather than the current 2 operations.

What do you think?

Flarna:
I guess `context.active()` as default in `getSpan`/`setSpan` would fit your needs, like this:
```ts
function getSpan(context = openTelemetry.context.active()): Span | undefined {
  return (context.getValue(SPAN_KEY) as Span) || undefined;
}
```

There are already places (e.g. `startSpan`) where similar defaulting is done instead of adding an extra API.

The only pitfall with such defaulting is that it relies on the global installed context manager. So if user prefers to use a non global one extra care is needed. But I think using non global `ContextManager`, `TracerProvider`,... is anyway exotic outside of tests.

cartermp:
That could probably work. Is there a formal or formal-ish API review process? I'm happy to contribute here but want to make sure the potential solution is considered thoroughly before adding to the public API.

Just some immediate pros/cons for either approach that come to mind:

`getSpan(context = openTelemetry.context.active())`/`setSpan(context = openTelemetry.context.active())`

Pros:

* Similarity with `startSpan` - knowing how any one of these works means you could know the pattern for interacting with the other two
* Less API surface area

Cons:

* Need to learn a default behavior that isn't necessarily apparent based on the name

`getCurrentSpan()`/`setAsCurrentSpan()`

 (choosing an arbitrary name for the inverse of `get` just for symmetry)

Pros:

* Likely obvious what it does based on the name; probably no need to understand default vs. non-default behavior

Cons:

* More API surface area
* Lacks symmetry with other parts of the API by not following a similar default/non-default pattern such as with `startSpan`

[...]

PR #163, opened 2022-06-21
feat: Add getActiveSpan to trace API

cartermp:
Based on discussion in https://github.com/open-telemetry/opentelemetry-js-api/issues/162 I'm opening this up. Both approached discussed in the issue are present -- it'd be great to get to consensus on if we want to do this, and if so, which approach to take.

My preference would be to have the additional `getActiveSpan` function, but it'd be great to hear what others think.

N.B. the implementations here may be suboptimal; this is mostly just to aid discussion

### DIFF
--- a/src/trace/context-utils.ts
+++ b/src/trace/context-utils.ts
@@ -36,7 +36,7 @@ const SPAN_KEY = createContextKey('OpenTelemetry Context Key SPAN');
 }
 
 /**
- * Gets the current span from the current context, if one exists.
+ * Gets the span from the current context, if one exists.
  */
 export function getCurrentSpan(): Span | undefined {
   const ctx = ContextAPI.getInstance().active();
--- a/src/trace/context-utils.ts
+++ b/src/trace/context-utils.ts
@@ -19,7 +19,7 @@ import { Context } from '../context/types';
 import { Span } from './span';
 import { SpanContext } from './span_context';
 import { NonRecordingSpan } from './NonRecordingSpan';
-import { ContextAPI } from '../api/context';
+import { context } from '..';
 
 /**
  * span key
@@ -39,8 +39,7 @@ const SPAN_KEY = createContextKey('OpenTelemetry Context Key SPAN');
  * Gets the span from the current context, if one exists.
  */
 export function getCurrentSpan(): Span | undefined {
-  const ctx = ContextAPI.getInstance().active();
-  return getSpan(ctx);
+  return getSpan(context.active());
 }
 
 /**
--- a/src/trace/context-utils.ts
+++ b/src/trace/context-utils.ts
@@ -19,7 +19,7 @@ import { Context } from '../context/types';
 import { Span } from './span';
 import { SpanContext } from './span_context';
 import { NonRecordingSpan } from './NonRecordingSpan';
-import { context } from '..';
+import { ContextAPI } from '../api/context';
 
 /**
  * span key
@@ -39,7 +39,7 @@ export function getSpan(context: Context): Span | undefined {
  * Gets the span from the current context, if one exists.
  */
 export function getCurrentSpan(): Span | undefined {
-  return getSpan(context.active());
+  return getSpan(ContextAPI.getInstance().active());
 }
 
 /**
--- a/src/trace/context-utils.ts
+++ b/src/trace/context-utils.ts
@@ -19,6 +19,7 @@ import { Context } from '../context/types';
 import { Span } from './span';
 import { SpanContext } from './span_context';
 import { NonRecordingSpan } from './NonRecordingSpan';
+import { ContextAPI } from '../api/context';
 
 /**
  * span key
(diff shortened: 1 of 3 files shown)


dyladan:
Great timing I was just looking at your issue this morning. I think a convenience API for users is a great idea and since it can be implemented entirely in the API by calling other APIs it shouldn't even need any SDK changes. I don't think we have a need for _both_ of these new APIs just to avoid confusing our users, but either of them likely provides enough value to justify.

Personally I would probably rather do `getCurrentSpan` because it does not require changing any existing API which users are already familiar with. Some users may expect `getSpan` to return `undefined` if passed an `undefined` context.

cartermp:
Agreed, I wouldn't want both either. Shoulda clarified in the PR text that I only want one as well. I also prefer `getCurrentSpan`, and anecdotally that is also what I've seen people ask for.

dyladan:
Moving my comment to the issue to keep discussion together.

dyladan:
You can go ahead and finish this up for review. I'll bring it up in the SIG meeting wednesday

dyladan:
Please add tests if you can. something like this should work

```typescript
const span = {} as any as span; // doesn't really matter if this is a real span
const ctx = setSpan(Context.ROOT_CONTEXT, span);
context.setGlobalContextManager({ active: () => ctx } as any as ContextManager);

const active = trace.getCurrentSpan();
assert.equals(active, span);
```

src/trace/context-utils.ts
```diff
@@ -30,10 +31,18 @@ const SPAN_KEY = createContextKey('OpenTelemetry Context Key SPAN');
...
-export function getSpan(context: Context): Span | undefined {
+ export function getSpan(context: Context): Span | undefined {
   return (context.getValue(SPAN_KEY) as Span) || undefined;
 }
 
+/**
+ * Gets the current span from the current context, if one exists.
+ */
+export function getCurrentSpan(): Span | undefined {
+  const ctx = ContextAPI.getInstance().active();
```
dyladan, line 42: I think you may be able to `import { context } from '..';` and use that here instead of `ContextAPI.getInstance()` which will save some small number of bytes when minified.

src/trace/context-utils.ts
```diff
@@ -30,10 +31,18 @@ const SPAN_KEY = createContextKey('OpenTelemetry Context Key SPAN');
  *
  * @param context context to get span from
  */
-export function getSpan(context: Context): Span | undefined {
+ export function getSpan(context: Context): Span | undefined {
   return (context.getValue(SPAN_KEY) as Span) || undefined;
 }
 
+/**
+ * Gets the current span from the current context, if one exists.
```
dyladan, line 39:
```suggestion
 * Gets the span from the current context, if one exists.
```

cartermp:
@dyladan will add a test as well. That's proving a bit more difficult (no mocha experience, weird type errors when importing certain things) but I'll give it a shot. If for some reason I can't figure it out soon I'll let you know.

cartermp:
Looks like this is the source of 4 new lint warnings and a new circular dependency, hence the linter failing:

> I think you may be able to `import { context } from '..';` and use that here instead of `ContextAPI.getInstance()` which will save some small number of bytes when minified.

Moving back to `ContextAPI.getInstance()` resolves this. Any preference on how to proceed?

Otherwise I think this is ready for review + tested.

blumamir (approved):
This is still being discussed in the relevant issue #163 , but if we choose to go with `getCurrentSpan` option - current implementation LGTM

legendecas (approved):

legendecas:
Just a side note: it will be great if you can prefix the title of the PR with the type of the changes, like "feat:" for this one, as documented at https://github.com/open-telemetry/opentelemetry-js/blob/main/CONTRIBUTING.md#conventional-commit.

vmarchaud (approved):

dyladan:
@cartermp if you can rename the method we can get this merged

cartermp:
Roger roger

legendecas:
Should this be a semver-minor or a semver-patch? IIUC, semver-minor makes it incompatible with SDKs not upgraded to the latest version of @opentelemetry/api.

cartermp:
I guess according to semver this should be a minor update:

> MINOR version when you add functionality in a backwards compatible manner

https://semver.org/

But since semver is just a lie that we all agree to tell one another, if there's precedent for adding backwards-compatible features as a patch, then that's not a bad option.

dyladan:
> I guess according to semver this should be a minor update:
> 
> > MINOR version when you add functionality in a backwards compatible manner
> 
> [semver.org](https://semver.org/) 

Technically yes. It would be a bummer for it to be incompatible because of version and not because of any actual incompatibility though.

> semver is just a lie that we all agree to tell one another

😂

We don't actually have any precedent to add features in patch releases but that doesn't mean I'm completely against it. I think the real problem here is we're trying to use one version number to represent the versions of 2 different things. One is the user-facing API and the other is the interface between the API and SDK. This change is minor for the first thing, but actually no change for the second thing. We might be better off not using the public version to track that compatiblity (might be too late to change though).

cartermp:
Since I'm not a maintainer, take my opinion with a grain of salt, but after sleeping on it I think a semver-minor change is the most appropriate, even though the api and SDK have this special relationship. I think it's a lot more likely that a developer will expect to get new functionality like this in a minor version bump than a patch, especially since it's the official guidance for package authorship that NPM gives too. And even though I'd love for people on older SDKs to get this functionality, they are far outnumbered by the people who will use otel-js for the first time over the next few years (and they will likely update over time too!).

naseemkullah (approved):

dyladan (approved):

PR #166, opened 2022-07-12
feat: deprecate Sampler

### LINKED ISSUE #148: `Sampler` and `SamplingResult` interfaces and implementations in SDK
As of now these interfaces are in `@opentelemetry/api` and implementation are part of `@opentelemetry/core`. Ideally they should have been in base sdk.

legendecas:
Preferring the one declared in @opentelemetry/sdk-trace-base.

Fixes: https://github.com/open-telemetry/opentelemetry-js-api/issues/148
Related: https://github.com/open-telemetry/opentelemetry-js/pull/3088

### DIFF
--- a/src/trace/Sampler.ts
+++ b/src/trace/Sampler.ts
@@ -21,6 +21,7 @@ import { SamplingResult } from './SamplingResult';
 import { SpanKind } from './span_kind';
 
 /**
+ * @deprecated use the one declared in @opentelemetry/sdk-trace-base instead.
  * This interface represent a sampler. Sampling is a mechanism to control the
  * noise and overhead introduced by OpenTelemetry by reducing the number of
  * samples of traces collected and sent to the backend.
--- a/src/trace/SamplingResult.ts
+++ b/src/trace/SamplingResult.ts
@@ -17,6 +17,7 @@
 import { SpanAttributes } from './attributes';
 
 /**
+ * @deprecated use the one declared in @opentelemetry/sdk-trace-base instead.
  * A sampling decision that determines how a {@link Span} will be recorded
  * and collected.
  */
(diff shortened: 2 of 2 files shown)


src/trace/Sampler.ts
```diff
@@ -21,6 +21,7 @@ import { SamplingResult } from './SamplingResult';
 import { SpanKind } from './span_kind';
 
 /**
+ * @deprecated use the one declared in @opentelemetry/sdk-trace-base instead.
```
Flarna, line 24: This might result in lint warning for users because a lot modules (e.g. all our instrumentations) are built on top of API 1.0.

Therefore users either have to disable the corresponding lint rule or accept the warning (like we do in SDK for Attributes).

legendecas: It's true that adding the `@deprecated` doc annotation might introduce either lint warnings or errors. But I don't think instrumentations are using the `Sampler` interface anywhere since it is not part of the API. And people using the `Sampler` interface from the new API package should have upgraded the SDK package. That's where we are trying to ask people to migrate to the declaration from SDK package right?

legendecas: When you use a newer version of API (like say v1.2.0), you can not use SDKs that aim lower version of API (v1.1.0) right? So when a person upgrades the API version, they'll have to upgrade the SDK version too, which should include their own declaration of `Sampler`.

dyladan: It's actually the opposite. You can upgrade your API all you want without a new SDK.

Flarna: The implementation of the sampler may be done independent of the application which chooses the final API/SDK. e.g. someone publishes a fancy sampler as an NPM package like it's done now with instrumentations, propagators,...

But I don't think it's blocking - it's just about a lint warning like we have it already with `Attributes` in SDK (and I believe others get it also).

dyladan: I think we shouldn't consider a lint warning breaking

legendecas:
> feat updates the minor version. Is that what we want?

Yes, this is intended. This is not breaking but introducing something new for SDKs to implement.

vmarchaud (approved):

dyladan (approved):

PR #172, opened 2022-08-24
feat(TraceProvider): add shutdown and forceFlush to API

secustor:
This PR exposes the `shutdown` and `forcedFlush` functions defined by the [TraceProvider specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#shutdown). 

Providing these methods allow to gracefully shutdown instrumentations without data loss. Currently this is only possible with usage of the SDK specific implementations [see](https://github.com/open-telemetry/opentelemetry-js/issues/3310)
## Context
Closes https://github.com/open-telemetry/opentelemetry-js/issues/3310

### DIFF
--- a/src/trace/NoopTracerProvider.ts
+++ b/src/trace/NoopTracerProvider.ts
@@ -33,4 +33,12 @@ export class NoopTracerProvider implements TracerProvider {
   ): Tracer {
     return new NoopTracer();
   }
+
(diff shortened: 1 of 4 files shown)


dyladan:
@secustor can you please re-apply this in https://github.com/open-telemetry/opentelemetry-js where this package is now moved?

Issue #165, opened 2022-07-06
Add a `withSpan` (or similar) convenience API


dyladan:
We can close this issue since the question is addressed and we can create new issues for proposed convenience APIs

_Originally posted by @dyladan in https://github.com/open-telemetry/opentelemetry-js-api/issues/164#issuecomment-1176500245_

I propose we add an API which takes a callback, automatically creating and ending a span. This has been proposed in the past and blocked for administrative reasons explained in the previously linked ticket, but I think now is the time to address it.

vmarchaud:
For reference, we made those utils at my company:
```ts
import * as otelAPI from '@opentelemetry/api'
import * as otelCore from '@opentelemetry/core'

export function withSpan <T extends () => ReturnType<T>> (span: otelAPI.Span, fn: T): ReturnType<T> {
  return otelAPI.context.with(
    otelAPI.trace.setSpan(otelAPI.context.active(), span),
    fn
  )
}

export function withoutSpan <T extends () => ReturnType<T>> (fn: T): ReturnType<T> {
  return otelAPI.context.with(
    otelCore.suppressTracing(otelAPI.context.active()),
    fn
  )
}

export function newSpan (spanName: string, attributes: otelAPI.SpanAttributes) {
  return otelAPI.trace.getTracer('default').startSpan(spanName, {
    attributes
  })
}

type Unpromisify<T> = T extends Promise<infer U> ? U : T
export async function withNewSpan <T extends () => ReturnType<T>> (
  spanName: string,
  attributes: otelAPI.SpanAttributes,
  fn: T
): Promise<Unpromisify<ReturnType<T>>> {
  const span = newSpan(spanName, attributes)
  // note: we await even if not needed to ensure we end() the span when the function
  // as been fully executed
  // tslint:disable-next-line: await-promise
  const result = await withSpan(span, fn)
  span.end()
  return result as any
}

export function withNewSpanSync <T extends () => ReturnType<T>> (
  spanName: string,
  attributes: otelAPI.SpanAttributes,
  fn: T
): ReturnType<T> {
  const span = newSpan(spanName, attributes)
  const result = withSpan(span, fn)
  span.end()
  return result as any
}
```

dyladan:
I wonder if `getTracer` should return a "sugared" tracer which implements convenience APIs so that we don't have to implement in SDK (and bump required version)?

```typescript
export class SugaredTracer implements Tracer {
  constructor(private _tracer: Tracer) {
[...]

PR #173, opened 2022-08-26
feat(SugaredTracer): add draft of sugaredTracer

secustor:
Early draft of a SugaredTracer, intended as basis for discussions and implementation ideas. 

cc @cartermp

Signed-off-by: secustor <sebastian@poxhofer.at>

### DIFF
--- a/src/trace/SugaredTracer.ts
+++ b/src/trace/SugaredTracer.ts
@@ -1,0 +1,229 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import type {Context, Span, SpanOptions, Tracer} from '../';
+import * as api from '../';
+import {SpanStatusCode} from '../';
+
+const defaultOnException = (e: Error, span: Span) => {
+  span.recordException(e)
+  span.setStatus({
+    code: SpanStatusCode.ERROR
+  })
+};
+
+/**
+ * return a new SugaredTracer created from the supplied one
+ * @param tracer
+ */
+export function wrapTracer(tracer: Tracer): SugaredTracer {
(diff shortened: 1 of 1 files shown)


cartermp (commented):
Overall I like the interface. Really my only question is if `withSpan` (where it's not marked a active) should exist as an option.

src/trace/SugaredTracer.ts
```diff
@@ -0,0 +1,229 @@
...
+ * Executes fn, returns results and runs onException in the case of exception to allow overwriting of error handling
+ * @param span
+ * @param onException
+ * @param fn
+ */
+function handleFn<F extends (span: Span) => ReturnType<F>>(span: Span, onException: ((e: Error, span: Span) => void), fn: F): ReturnType<F> {
+  try {
+    const ret = fn(span);
+    // if fn is an async function attach a recordException and spanEnd callback to the promise
+    if (ret instanceof Promise) {
```
legendecas, line 215: We don't generally brand check Promise in JavaScript. Typically, a duck-type checking like `typeof ret.then === 'function'` is preferred for thanables.

src/trace/SugaredTracer.ts
```diff
@@ -0,0 +1,229 @@
...
+
+/**
+ * return a new SugaredTracer created from the supplied one
+ * @param tracer
+ */
+export function wrapTracer(tracer: Tracer): SugaredTracer {
+  return new SugaredTracer(tracer)
+}
+
+class SugaredTracer implements Tracer {
```
legendecas, line 35: Bikeshedding: I'm not sure if this name is appropriate or comprehensible to users. But I didn't come up with a better name :(

legendecas (commented):
I like the idea of creating a new wrapping interface that does not require SDK to provide any extra interface implementations!

src/trace/SugaredTracer.ts
secustor: I have based the current iteration on the discussion in [this issue]( https://github.com/open-telemetry/opentelemetry-js-api/issues/165), but I'm open for alternatives :)

cartermp: How about `ConvenientSpanTracer` or `AutoClosingSpanTracer`?

secustor: `AutoClosingSpanTracer`: IMO this will not include only this feature, but could be expanded in the future. Therefore I think we should not use something specific like this.     
`ConvenientSpanTracer`: Is fitting but, `Sugared` seems to be widely used for such scenarios. 
- https://github.com/golang/go/discussions/54763#discussioncomment-3508573
- https://github.com/uber-go/zap#quick-start

dyladan:
@secustor this package is being moved into the main repo. Can you please re-apply this change there? I realize it isn't ideal, but it should at least get more attention there.

PR #174, opened 2022-09-07
Export NoopTracer

danilobuerger:
I would like to use this to opt out of tracing under certain conditions.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -23,6 +23,8 @@ export * from './diag';
 export * from './propagation/TextMapPropagator';
 export * from './trace/attributes';
 export * from './trace/link';
+export * from './trace/NoopTracer';
+export * from './trace/NoopTracerProvider';
 export * from './trace/ProxyTracer';
 export * from './trace/ProxyTracerProvider';
 export * from './trace/Sampler';


Flarna:
Could you provide some more details here?
Usually samplers are the entities to decide if an exported span should be created or not.

To completely opt out from tracing the easiest way is to just not call `setTracerProvider`.

danilobuerger:
It would make it inline with other language sdks, for example go exposes it: https://pkg.go.dev/go.opentelemetry.io/otel/trace#NewNoopTracerProvider

Flarna:
@danilobuerger That doesn't answer my questions. One could argue it should be removed there :o)

danilobuerger:
Thats fine. Not here to argue.

Flarna:
The main reason for not exporting is to reduce the amount of APIs to maintain and to avoid wrong/misuse. A similar API cleanup was done for metrics to prepare it for GA.

Therefore I asked for the use case. If it is a valid use case which can't be archived with current API exposing the Noop classes/instances might be good. Or to add another API to fulfill your usecase.

But I see no reason to expose them just to have them exposed.

danilobuerger:
For my use case I already applied this PR with patch-package so its already working for me and I am using the `NoopTracer`. I just thought I would surface this upstream here. If it's not wanted, thats fine.

dyladan:
Moving discussion of this request to https://github.com/open-telemetry/opentelemetry-js/issues/3315 as this repo is being archived.

PR #175, opened 2022-09-13
feat: never return undefined when the method expects an object

dmathieu:
I believe returning `undefined` when the context has no span is invalid per the specification:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#guidance

> Whenever API call returns values that is expected to be non-null value - in case of error in processing logic - SDK MUST return a "no-op" or any other "default" object that was (ideally) pre-allocated and readily available. This way API call sites will not crash on attempts to access methods and properties of a null objects.

See also the discussion in https://github.com/open-telemetry/oteps/issues/216

Note: this may be deemed a breaking change, in which case I am happy to rename the commit to `feat`, and to wait for the next planned appropriate release.

### DIFF
--- a/src/baggage/types.ts
+++ b/src/baggage/types.ts
@@ -61,7 +61,7 @@ export interface Baggage {
    *
    * @param key The key which identifies the BaggageEntry
    */
-  getEntry(key: string): BaggageEntry | undefined;
+  getEntry(key: string): BaggageEntry;
 
   /**
    * Get a list of all entries in the Baggage
--- a/src/baggage/context-helpers.ts
+++ b/src/baggage/context-helpers.ts
@@ -17,6 +17,7 @@
 import { createContextKey } from '../context/context';
 import { Context } from '../context/types';
 import { Baggage } from './types';
+import { createBaggage } from './utils';
 
 /**
  * Baggage key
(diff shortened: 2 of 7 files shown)


Flarna (changes requested):
marking with request change until the spec part is clear regarding this.

If we go this root we should not stick on `getSpan` instead we should fix this for all context methods (e.g. baggage,...).

dmathieu:
I have added this to every other public method which returned undefined (and not a string). This is way farther from my area of confidence in the JS API though. So I may have messed up.

Not having this is currently forcing us into writing a lot of defensive conditions. Basically, whenever we retrieve the current span from context, we need to check if there is indeed a span, which is quite verbose.

src/baggage/types.ts
```diff
@@ -61,7 +61,7 @@ export interface Baggage {
    *
    * @param key The key which identifies the BaggageEntry
    */
-  getEntry(key: string): BaggageEntry | undefined;
```
legendecas, line 64: This removes the ability to determine if a baggage entry is present in the baggage. I'd find this line of change unexpected regarding the OP: 

> Whenever API call returns values that is expected to be non-null value - in case of error in processing logic

The `undefined` returned here is not an error.

dyladan:
I don't think returning undefined is invalid.

>Whenever API call returns values **that is expected to be non-null value** - in case of error in processing logic - SDK MUST return a "no-op" or any other "default" object that was (ideally) pre-allocated and readily available. This way API call sites will not crash on attempts to access methods and properties of a null objects.

IMO, this method is not expected to return a non-null value. The docs explicitly state that it may be undefined. Further, this spec guidance is specifically around error handling and what you've changed is the normal operation of the API. This change could also be interpreted as breaking.

This package is being moved to the main https://github.com/open-telemetry/opentelemetry-js repository and I would suggest that if you feel this is an important issue you open an issue there.

dyladan:
I went ahead and created a discussion issue over there https://github.com/open-telemetry/opentelemetry-js/issues/3316

